PR #32, opened 2025-02-28
update path

youkaichao:
`./insmod.sh` is at the same level as `packages`


youkaichao:
see https://github.com/NVIDIA/gdrcopy/tree/master

PR #66, opened 2025-03-10
Allow passing output tensor in low_latency_combine

dzhulgakov:
It's convenient for explicit buffer management in some situations. The convention with optional `out=` arguments matches PyTorch common practices.

### DIFF
--- a/deep_ep/buffer.py
+++ b/deep_ep/buffer.py
@@ -498,7 +498,7 @@ class Buffer:
     # noinspection PyTypeChecker
     def low_latency_combine(self, x: torch.Tensor, topk_idx: torch.Tensor, topk_weights: torch.Tensor,
                             handle: tuple, async_finish: bool = False, return_recv_hook: bool = False,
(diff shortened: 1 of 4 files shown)


deep_ep/buffer.py
```diff
@@ -497,7 +497,8 @@ def low_latency_dispatch(self, x: torch.Tensor, topk_idx: torch.Tensor,
 
     # noinspection PyTypeChecker
     def low_latency_combine(self, x: torch.Tensor, topk_idx: torch.Tensor, topk_weights: torch.Tensor,
-                            handle: tuple, async_finish: bool = False, return_recv_hook: bool = False) -> \
+                            handle: tuple, async_finish: bool = False, return_recv_hook: bool = False,
+                            out: torch.Tensor | None = None) -> \
```
LyricZhao, line 501: It is not compatible with Python 3.8, can you change it to `Optional[torch.Tensor]`?

PR #96, opened 2025-03-28
Adjust kNumThreads of notify_dispatch

songhexiang:
For the SMs which calculate metadata in notify_dispatch, each warp in the SM is used to calculate the metadata of one channel.
The default configuration is **8 warps for 10 channels**, which needs two rounds of loop. 
Maybe the number of warps can be configured to **the number of the channels** so that one loop is enough.

### DIFF
--- a/csrc/kernels/internode.cu
+++ b/csrc/kernels/internode.cu
@@ -428,7 +428,7 @@ void notify_dispatch(const int* num_tokens_per_rank, int* moe_recv_counter_mappe
                   buffer_ptrs, task_fifo_ptrs, head, rank, \
                   cpu_rdma_team); } break
 
-    constexpr int kNumThreads = 256;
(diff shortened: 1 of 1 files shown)


LyricZhao:
Here the `kNumThreads` is a constexpr, changed it to 512, see https://github.com/deepseek-ai/DeepEP/commit/c4d12b4f8f322c75862aa644a5ef74fac661005a. Thanks for your feedback! (but actually the bottleneck of this kernel is SM 0, boosting other SMs may only save some GPU power and may have no impact on performance)

songhexiang:
Sorry, I did not check that the num_channels is not a constant variable
Indeed the bottleneck is in SM0.

LyricZhao:
> Indeed the bottleneck is in SM0.

Yep, but saving power is also meaningful, right? :)

PR #130, opened 2025-04-22
Support multi-QP for normal kernels

LyricZhao:
This PR is authored by **Tencent Network Platform Department**. Thanks for the contribution! Now normal kernels have a huge speedup:

|   Type    | Dispatch #EP | Bottleneck bandwidth | Combine #EP | Bottleneck bandwidth |
|:---------:|:------------:|:--------------------:|:-----------:|:--------------------:|
| Internode |      32      |    44 👉🏻 58 GB/s (RDMA)    |     32      |    47 👉🏻 57 GB/s (RDMA)    |
| Internode |      64      |    46 👉🏻 51 GB/s (RDMA)    |     64      |    45 👉🏻 50 GB/s (RDMA)    |

Through in-depth optimization, the following enhancements have been implemented in the internode normal kernels.

- Replacing IBRC with IBGDA
- Utilizing distinct QPs (Queue Pairs) per channel for parallel data transmission

These improvements not only enhance the robustness of the Internode Normal Kernel in scenarios involving dual-port NICs and RoCE networks but also further elevate communication performance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,11 +17,11 @@ We test normal kernels on H800 (~160 GB/s NVLink maximum bandwidth), with each c
 |   Type    | Dispatch #EP | Bottleneck bandwidth | Combine #EP | Bottleneck bandwidth |
 |:---------:|:------------:|:--------------------:|:-----------:|:--------------------:|
 | Intranode |      8       |  153 GB/s (NVLink)   |      8      |  158 GB/s (NVLink)   |
-| Internode |      16      |    47 GB/s (RDMA)    |     16      |    62 GB/s (RDMA)    |
-| Internode |      32      |    59 GB/s (RDMA)    |     32      |    60 GB/s (RDMA)    |
-| Internode |      64      |    49 GB/s (RDMA)    |     64      |    51 GB/s (RDMA)    |
+| Internode |      16      |    43 GB/s (RDMA)    |     16      |    43 GB/s (RDMA)    |
+| Internode |      32      |    58 GB/s (RDMA)    |     32      |    57 GB/s (RDMA)    |
+| Internode |      64      |    51 GB/s (RDMA)    |     64      |    50 GB/s (RDMA)    |
 
-**News (2025.04.22)**: the performance is optimized by 5-35% by Tencent Network Platform Department, see [#130](https://github.com/deepseek-ai/DeepEP/pull/130) for more details. Thanks for the contribution!
+**News (2025.04.22)**: with optimizations from Tencent Network Platform Department, performance was enhanced by up to 30%, see [#130](https://github.com/deepseek-ai/DeepEP/pull/130) for more details. Thanks for the contribution!
 
 ### Low-latency kernels with pure RDMA
 
--- a/README.md
+++ b/README.md
@@ -21,6 +21,16 @@ We test normal kernels on H800 (~160 GB/s NVLink maximum bandwidth), with each c
 | Internode |      32      |    44 GB/s (RDMA)    |     32      |    47 GB/s (RDMA)    |
 | Internode |      64      |    46 GB/s (RDMA)    |     64      |    45 GB/s (RDMA)    |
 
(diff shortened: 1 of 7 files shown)


ethnzhng:
Hi, thanks for the awesome open-source work. For the new normal internode kernels, do we have insight into the reason why the bandwidth decreases from EP32 --> 64, inversely to the increase from EP16 --> 32? Or was the bandwidth for EP16 not updated in the table?

sphish:
@ethnzhng
1. For the EP16 case: the performance is almost the same on our H800 server. However, the test results from the Tencent team on the H20 server demonstrated a significant performance improvement. I think it is because the NVLink on H800 has become the bottleneck in this test case.
2. For the bandwidth decreasing from EP32 -> 64:  the test results shown here represent algorithmic bandwidth. In reality, for the EP32 case, 1/4 of the data doesn't go through the RDMA network, so the physical bandwidth should be multiplied by 3/4. Similarly, for EP64, the physical bandwidth should be multiplied by 7/8. If you compare using physical bandwidth, you'll find they are essentially the same.

18018681653:
I found that although the low latency kernel has multiple QPS, due to the imbalance problem of the moe itself, the amount of data sent by each qp is also imbalance. Therefore, the performance under dual-port is worse than that under single-port. Do you have a solution

whybeyoung:
LGTM

sphish:
@18018681653 Since we don't use dual-port NICs ourselves, we currently won't be adapting to dual-port support. The good news is that I asked the Tencent team @moningchen , and they may open-source their dual-port solution in the future. You can keep an eye on them!

GHGmc2:
May I know why can we get higher bandwidth than 50GB/s since CX7 IB NIC has only 400Gbps(50GB/s)?

sphish:
@GHGmc2 As mentioned above, what we report is the algorithmic bandwidth, not the physical bandwidth.

yiakwy-xpu-ml-framework-team:
> @ethnzhng
> 
> 1. For the EP16 case: the performance is almost the same on our H800 server. However, the test results from the Tencent team on the H20 server demonstrated a significant performance improvement. I think it is because the NVLink on H800 has become the bottleneck in this test case.
> 2. For the bandwidth decreasing from EP32 -> 64:  the test results shown here represent algorithmic bandwidth. In reality, for the EP32 case, 1/4 of the data doesn't go through the RDMA network, so the physical bandwidth should be multiplied by 3/4. Similarly, for EP64, the physical bandwidth should be multiplied by 7/8. If you compare using physical bandwidth, you'll find they are essentially the same.

Thanks for clarifying the computation issue. I think it better to udpate the table with device name , bandwidth name (bus bandwidth or algorithm bandwith).

yiakwy-xpu-ml-framework-team:
@LyricZhao , @sphish  in normal configuration, we have GPU:NIC = 1:1, so with 400 Gb network card support, we will have aggreated 3200 Gb commnication bandwidth. Is it true for your testing configuration ?

MaoZiming:
@LyricZhao Hi, does this PR replace IBRC with IBGDA (?) It seems IBGDA was enabled before the PR https://github.com/deepseek-ai/DeepEP/pull/130/files#diff-a2699bac7470d7e4b1c4458ed71afe7bc3bae2970f271e5a10c9b4bb9cbde223L73

sphish:
> @LyricZhao , @sphish in normal configuration, we have GPU:NIC = 1:1, so with 400 Gb network card support, we will have aggreated 3200 Gb commnication bandwidth. Is it true for your testing configuration ?

@yiakwy-xpu-ml-framework-team Yes.

fengidri:
Please, can someone explain this improvement - what role did multiqueue and ibgda each play? My personal understanding is that ibgda has more advantages in terms of latency, so what is the reason for completely switching to ibgda? I haven't found an explanation for this point. If I missed something, please point it out.

ztxdcyy:
“Utilizing distinct QPs (Queue Pairs) per channel for parallel data transmission” Can anyone explain what's the meaning of channel? or where can i find it? Thanks a lot!!

alpha-baby:
> “Utilizing distinct QPs (Queue Pairs) per channel for parallel data transmission” Can anyone explain what's the meaning of channel? or where can i find it? Thanks a lot!!


https://github.com/deepseek-ai/DeepEP/blob/main/csrc/config.hpp#L89


In DeepEP, "channel" is a concept that divides different SMs (Streaming Multiprocessors) into multiple groups, with two SMs in each group. We assign a number to each group. For example: the first group is called channel-0, the second group is called channel-1, and so on.

I suspect that the concept of "channel" originates from NCCL.

PR #142, opened 2025-04-29
Fix DeepEP cannot be used together with code that needs GIL such as Mooncake transfer engine

fzyzcjy:
When using DeepEP with Mooncake transfer engine, DeepEP's dispatch will hold the GIL for a long time, and thus other libraries such as Mooncake transfer engine, which uses Python code in another thread, cannot obtain the GIL and thus get stuck. Therefore, in this PR, I make DeepEP release GIL to fix this.


alpha-baby:
in sglang code:
image

If it is asynchronous mode, it is reasonable to say that this function should not hold CIL for a long time.

PR #151, opened 2025-05-09
Feat: enhance nvidia peer memory detection

vicoooo26:
In my H800 environment which NVIDIA Driver is `550.54.15` and `nvidia_peermem` module has been loaded successfully, but NVSHMEM detections of `nvidia_peermem` existence never pass, then I will get some error messages like `neither nv_peer_mem, or nvidia_peermem detected. Skipping transport.`. That's because there is not neither `/sys/kernel/mm/memory_peers/nv_mem/version` nor `/sys/kernel/mm/memory_peers/nv_mem_nc/version` file in my machine.
According to [NCCL's implementation](https://github.com/NVIDIA/nccl/blob/3000e3c797b4b236221188c07aa09c1f3a0170d4/src/transport/net_ib.cc#L711) I use `/sys/module/nvidia_peermem/version` file as an indication of the existence of `nvidia_peermem`, then I can run DeepDP `internode`、`intronode` and `low_latency` over multiple nodes successfully after patching it.

### DIFF
--- a/third-party/nvshmem.patch
+++ b/third-party/nvshmem.patch
@@ -421,3 +421,17 @@ index ea1e284..e6640d6 100644
 --
 2.25.1
 
+diff --git a/src/modules/transport/common/transport_ib_common.cpp b/src/modules/transport/common/transport_ib_common.cpp
+index c89f408..f99018a 100644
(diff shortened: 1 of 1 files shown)


sphish:
Ok. I'll also share this feedback with the NVSHMEM team.

PR #161, opened 2025-05-13
support fp8 intranode_combine

Went-Liang:
support fp8 intranode_combine


LyricZhao:
Sorry, we won't merge it as we also have internal FP8 combine impl, and it is hard to make the performance good as there are lots of other operations (e.g. dequant).

For model training consideration, FP8 combine will cause model's accuracy on benchmarks lower, even for inference only. We also have some other quant strategies, e.g. dynamic FP8 cast/customized non-FP8 fmts internally, but none of them offers both good performance and good model accuracy.

So quantization of `combine` is sensitive, we may add such features later with careful verification.

PR #191, opened 2025-06-06
Use TMA instead of LD/ST for intra-node normal kernels

LyricZhao:
Better performance for intranode EP kernels on H800/H20.


whybeyoung:
We tested it on H20 and didn't feel a significant gain. Do we need any specific configurations? @LyricZhao 
prefill (ep16 dp2)

alpha-baby:
> We tested it on H20 and didn't feel a significant gain. Do we need any specific configurations? @LyricZhao prefill (ep16 dp2)

this PR only  support intra-node,

EP16 should try this: https://github.com/deepseek-ai/DeepEP/pull/276#event-18462625428

FireKingY:
Hello, I'd like to ask why the sender doesn't use TMA? I noticed in the commit history that it was initially changed to have the sender use TMA, but then later reverted

PR #195, opened 2025-06-09
Add low-latency kernel PCIe usage flag

LyricZhao:


sphish (approved):

alpha-baby:
Is there any specific usage scenario for the flag added in this PR? I don't quite understand. Can you give me a simple example?

LyricZhao:
KVCache transfer requires H2D/D2H, which occupies PCIe bandwidth (EP also uses PCIe to network card). So the KVCache transfer CPU thread should check this flag and find the non-EP time to do KV H2D/D2H.

alpha-baby:
> KVCache transfer requires H2D/D2H, which occupies PCIe bandwidth (EP also uses PCIe to network card). So the KVCache transfer CPU thread should check this flag and find the non-EP time to do KV H2D/D2H.

https://github.com/deepseek-ai/DeepEP/pull/214
why remove this feature?

LyricZhao:
As far as I know, nobody in the open-source communitiy uses such feature.

alpha-baby:
> As far as I know, nobody in the open-source communitiy uses such feature.

i got, thank you for your reply.

PR #201, opened 2025-06-10
remove the dependency of gdrcopy

youkaichao:
Right now we use either nvlink or ibgda, gdrcopy is not needed anymore.

I find that modifying the `nvshmemt_ibgda_progress` function can make nvshmem compile without gdrcopy. This can greatly simplify the process of installing DeepEP. We just need to compile nvshmem, with one additional driver flag `options nvidia NVreg_EnableStreamMemOPs=1 NVreg_RegistryDwords="PeerMappingOverride=1;"` .

NOTE: theoretically ibgda has a cpu-assisted mode that needs gdrcopy, however, the default mode is "auto", and if we compile without gdrcopy, the cpu-assisted mode will be turned off automatically. therefore, we don't need to do anything special in deepep.

### DIFF
--- a/third-party/README.md
+++ b/third-party/README.md
@@ -8,66 +8,19 @@
 
 ## Prerequisites
 
-1. [GDRCopy](https://github.com/NVIDIA/gdrcopy) (v2.4 and above recommended) is a low-latency GPU memory copy library based on NVIDIA GPUDirect RDMA technology, and *it requires kernel module installation with root privileges.*
-
-2. Hardware requirements
-   - GPUDirect RDMA capable devices, see [GPUDirect RDMA Documentation](https://docs.nvidia.com/cuda/gpudirect-rdma/)
+Hardware requirements:
+   - GPUs inside one node needs to be connected by NVLink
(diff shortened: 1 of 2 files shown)


third-party/README.md
```diff
@@ -8,74 +8,27 @@
 
 ## Prerequisites
 
-1. [GDRCopy](https://github.com/NVIDIA/gdrcopy) (v2.4 and above recommended) is a low-latency GPU memory copy library based on NVIDIA GPUDirect RDMA technology, and *it requires kernel module installation with root privileges.*
-
-2. Hardware requirements
-   - GPUDirect RDMA capable devices, see [GPUDirect RDMA Documentation](https://docs.nvidia.com/cuda/gpudirect-rdma/)
+Hardware requirements:
+   - GPUs inside one node needs to be connected by NVLink
+   - GPUs across different nodes needs to be connected by RDMA devices, see [GPUDirect RDMA Documentation](https://docs.nvidia.com/cuda/gpudirect-rdma/)
```
sphish, line 13: In the low latency mode, all GPUs should be connected with RDMA.

sphish: I forgot low latency kernels can leverage NVLink now.

PR #204, opened 2025-06-11
Support Ampere architecture

LyricZhao:
Please notice that this support is for intranode only.


deep_ep/utils.py
```diff
@@ -58,3 +61,28 @@ def __exit__(self, exc_type: Any, exc_val: Any, exc_tb: Any) -> None:
         """
         if self.event is not None:
             self.event.current_stream_wait()
+
+
+def check_nvlink_connections(group: dist.ProcessGroup):
```
youkaichao, line 66: FYI vllm uses `pynvml` to check the nvlink connectivity:

https://github.com/vllm-project/vllm/blob/b8e809a057765c574726a6077fd124db5077ce1f/vllm/platforms/cuda.py#L438

LyricZhao: Thanks, I will try it later!

LyricZhao: See #205.

PR #205, opened 2025-06-11
Use `pynvml` to detect NVLink connections

LyricZhao:


deep_ep/utils.py
```diff
@@ -73,16 +73,25 @@ def check_nvlink_connections(group: dist.ProcessGroup):
     # Check NVLink connection
     # NOTES: some A100 PCIE GPUs only have pairwise NVLink connection, so that we can only use EP2
     if 'PCIE' in torch.cuda.get_device_name():
-        assert group.size() <= 2, 'No NVLink connection between all GPUs'
+        assert group.size() <= 2, 'PCIe GPUs only have pairwise NVLink connections'
+
+        # noinspection PyUnresolvedReferences
+        import pynvml
+        pynvml.nvmlInit()
```
youkaichao, line 81: need to call `pynvml.nvmlShutdown()` when the function returns.

LyricZhao:
> need to add https://pypi.org/project/nvidia-ml-py/ into dependency?

This repo does not maintain `requirements.txt`. And currently the `import` only occurs with PCIe A100s. So I guess this is OK now.

PR #238, opened 2025-06-20
Tiny support tests on systems with other number of GPUs

fzyzcjy:
Some systems do not have exactly 8 gpus


fzyzcjy:
Yes that looks great, but it seems DeepEP's tests all use env variables, thus I follow their convention. I am ok to make it argparse and/or refactor all other env-based code.

PR #239, opened 2025-06-21
Support displaying separate send and recv time

fzyzcjy:
This is useful in e.g.

* Want to use UT as a reference to check whether the e2e system behaves correctly for the send and recv kernel
* Want to speedup the send and recv kernels separately

Example output

image


LyricZhao:
Did some cleanup on https://github.com/deepseek-ai/DeepEP/commit/d79b3cd12f6a4e911d17492539da68b11e1a255d.

PR #245, opened 2025-06-23
Add the transaction window data structure for RDMA senders

LyricZhao:


sphish (approved):

yhyang201:
Hi, I have a quick question about the change in `csrc/kernel/s/launch.cuh`. I noticed that support for rank 3 (`case 3n: case_macro(3);`) was removed from the `SWITCH_ROW_RANKS` macro, and I was wondering what the reasoning behind this was. Any context you could provide would be much appreciated. Thanks!

LyricZhao:
Nothing special, 3 is a rare case and we did it to reduce the compilation time. If you have demand for this, you can submit a PR 👍🏻

PR #264, opened 2025-06-27
support hidden size 8192

ruizhang1230:


LyricZhao:
Thanks for your contribution!

PR #277, opened 2025-07-03
Allow using few SMs for low-latency mode

fzyzcjy:
The code diff is surely not for merging, but for demonstration how the experiments below are done. If anyone is interested / this direction looks acceptable to be merged, I am happy to polish and further work on the code!

The code and experiment data are extracted from old experiments for my previous https://github.com/deepseek-ai/DeepEP/pull/249.

Figure 1: num-sm vs performance
As can be seen, when using 9 warpgroup - ie few SMs, the performance only slightly slow down. Thus this makes a simple overlapping between this and computation feasible.

image

For dispatch we may need to do extra work though, since the warp specialization may be suboptimal when there are few SMs.

### DIFF
--- a/csrc/kernels/internode_ll.cu
+++ b/csrc/kernels/internode_ll.cu
@@ -339,7 +339,9 @@ void dispatch(void* packed_recv_x, void* packed_recv_x_scales,
               void* workspace, int num_device_sms,
               cudaStream_t stream, int phases) {
     constexpr int kNumMaxTopK = 9;
-    const int num_warp_groups = ceil_div(num_experts, num_device_sms);
+    const int num_warp_groups = ((phases & LOW_LATENCY_RECV_PHASE) == 0)
+        ? 9
+        : ceil_div(num_experts, num_device_sms);
     const int num_warps_per_group = 32 / num_warp_groups;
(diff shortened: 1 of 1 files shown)


rubbberrabbit:
> As can be seen, when using 9 warpgroup - ie few SMs, the performance only slightly slow down. Thus this makes a simple overlapping between this and computation feasible.

Hi, that is a good idea, but I want to understand the mentioned “overlap” in more detail. Does it refer to the overlap between the dispatch/combine kernels and the model computation kernels—i.e. two separate streams like in prefill? But during decode a CUDA Graph is enabled; wouldn’t the CUDA Graph turn them into a single sequential execution stream and eliminate that overlap?

alpha-baby:
Deepep kernel uses less SM, so who will use the extra SM? For example, in the decode phase of sglang, the communication kernel and the compute kernel are serial.

rubbberrabbit:
> Deepep kernel uses less SM, so who will use the extra SM? For example, in the decode phase of sglang, the communication kernel and the compute kernel are serial.

I think maybe it involves two batch overlap，so the the communication kernel and the compute kernel can run at the same time ideally. But still as I mentioned，after with CUDA graph，it transfer to serial kernel. I am very curious how to avoid it.

PR #284, opened 2025-07-08
Support 10-bit LogFMT (simulated version)

LyricZhao:


fzyzcjy:
This looks great and looking forward to it, 16bit ll combine is too slow currently even if it is near 700GB/s transfer speed and thus make naive communication computation overlap a bit hard

LyricZhao:
@fzyzcjy Thanks! We are merging this simulated branch now, and the real compressed version will be another PR. cc @zheanxu

fzyzcjy:
Hi, may I know when is the rough ETA about the real version? In addition, is this feature meant to be directly used in normal cases, i.e. DeepSeek decode inference, without accuracy loss, or is it to be used in some special cases and will loss accuracy in other cases. Thanks!

LyricZhao:
By the end of this week.

It is just for decoding. For normal kernels, quant/dequant costs a lot of CUDA cores, which will bound the communication bandwidth.

fzyzcjy:
I see, looking forward to it

> It is just for decoding.

Sure, I only use low_latency mode in decode

fzyzcjy:
> By the end of this week.

Hi, I would appreciate it if I could know the updated ETA

LyricZhao:
Sorry, we have something more urgent internally, we will continue to work on this after that. So the ETA is currently unknown :)

fzyzcjy:
Oh I see... anything I can help?

Btw, firstly to be sure: I would like to try this on some hardware, where the communication is purely via nvlink, and thus the time spent for moe gemm and deepep is of similar duration. Thus for computation communication overlap, it seems good to overlap the gemm and the deepep, instead of two-batch overlap. Thus DeepEP combine send kernel will be run using ~30 SM (preferably less). In this case, I wonder whether the 10bit format will still be faster than standard 16bit or not? (I originally thought casting is not expensive and kernel is still mem bound, but see the "For normal kernels, quant/dequant costs a lot of CUDA cores, which will bound the communication bandwidth." and thus need to be sure)

If this does speedup (a lot), then I am happy to contribute

LyricZhao:
> In this case, I wonder whether the 10bit format will still be faster than standard 16bit or not?

Sorry, we are still working on this, I am not sure how many SMs are needed. But I am sure this is very useful for large EP with RDMA.

LyricZhao:
@zheanxu  almost finished it, but debuging with some numeric error in some cases. I guess no need for help :) Thanks for your kindness.

fzyzcjy:
OK I see, then eagerly looking forward to it, and feel free to ping me if need help in other cases!

fzyzcjy:
@LyricZhao By the way, I would appreciate it if the partial code could be a (draft) PR, then I may be able to firstly test its performance with few SMs etc.

PR #292, opened 2025-07-10
Explicitly destroy the C++ runtime and release resources.

sphish:
Since operations like `nvshmem_finalize` may cause Python's exception handling process to get stuck, we decoupled the destructor and resource release in a manner similar to PyTorch NCCL.

### DIFF
--- a/csrc/deep_ep.cpp
+++ b/csrc/deep_ep.cpp
@@ -81,40 +81,10 @@ Buffer::Buffer(int rank, int num_ranks, int64_t num_nvl_bytes, int64_t num_rdma_
 }
 
 Buffer::~Buffer() noexcept(false) {
-    // Synchronize
-    CUDA_CHECK(cudaDeviceSynchronize());
-
-    if (num_nvl_bytes > 0) {
-        // Barrier
-        intranode::barrier(barrier_signal_ptrs_gpu, nvl_rank, num_nvl_ranks, comm_stream);
-        CUDA_CHECK(cudaDeviceSynchronize());
-
-        // Close remote IPC
-        if (is_available()) {
-            for (int i = 0; i < num_nvl_ranks; ++ i) if (i != nvl_rank)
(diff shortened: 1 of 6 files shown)


shifangx:
Thanks for @sphish 's fix pr and briefly introducing the root cause. The following is detailed explanation of this best practice.

### Why should explicitly decoupling destruction and cleanup

When integrating CUDA and multi-node communication libraries such as NVSHMEM into Python or C++ projects—especially those that combine Python object lifecycles and complex GPU resource management—it's crucial to separate object destruction from resource cleanup functions to avoid system hang issues. If heavy resource release (like nvshmem_finalize) happens automatically in destructors (~Buffer()), unexpected interactions with Python’s garbage collection or exception-handling process can cause deadlocks. For instance, Python's interpreter may be in a state where global communication or memory synchronization is unsafe, especially if there are lingering references or if finalization triggers as part of error handling. This scenario is particularly risky in deep learning frameworks where both C++ and Python manage GPU and distributed resources.

### How does explicitly decoupling destruction and cleanup works

By explicitly decoupling destruction and cleanup—having an independent destroy() method—the programmer can control exactly when and in what context heavy resource release functions are safely called. This helps ensure such calls aren't made while Python is handling exceptions or shutting down, which can conflict with lower-level system states and hang processes. This approach echoes best practices in resource-intensive frameworks like PyTorch and NCCL, which caution against automating global finalize steps in destructors. Instead, they recommend explicit release methods, allowing developers to clean up resources in controlled, predictable contexts, away from Python exception or shutdown logic.

### Explicit warnings are also useful

Explicit warnings when users forget to call these cleanup methods before destruction are also valuable—they signal the risk of resource leaks, but crucially avoid unsafe finalize calls that could deadlock processes. In summary, explicit resource management through decoupled cleanup functions gives more predictable lifecycles, safer finalization, and helps prevent hangs in high-performance, multi-process CUDA Python environments.

PR #293, opened 2025-07-10
support low latency dispatch tma

ayrnb:
A practice to learn the use of TMA. If there are any errors here, hope to get your guidance and feedback. :pray::pray::pray::pray:


fzyzcjy:
Hi, curious how much performance improvement does it have?

hideonevan:
Why set the mbarrier's expected arrival to 1，But a warp's threads all mbarrier_arrive

PR #295, opened 2025-07-12
third-party: Improvements to NVSHMEM Integration

seth-howell:
1. Use upstream NVSHMEM binaries when building DeepEP.
2. Add back support for CPU-Assisted IBGDA.
3. Remove the nvshmem host-side patch.

### DIFF
--- a/third-party/README.md
+++ b/third-party/README.md
@@ -52,7 +52,7 @@ sudo reboot
 
 #### 2.2 Install GDRCopy and load the gdrdrv kernel module
 
-This configuration enables IBGDA through asynchronous post-send operations assisted by the CPU.
+This configuration enables IBGDA through asynchronous post-send operations assisted by the CPU. More information about CPU-assisted IBGDA can be found in [this blog](https://developer.nvidia.com/blog/enhancing-application-portability-and-compatibility-across-new-platforms-using-nvidia-magnum-io-nvshmem-3-0/#cpu-assisted_infiniband_gpu_direct_async%C2%A0).
 It comes with a small performance penalty, but can be used when modifying the driver regkeys is not an option.
 
 Download GDRCopy


third-party/nvshmem.patch
youkaichao, line 1: please keep this file right now since vllm's installation guide still uses it. we will catch up later.

third-party/README.md
```diff
@@ -45,38 +50,19 @@ sudo update-initramfs -u
 sudo reboot
 ```
 
-For more detailed configurations, please refer to the [NVSHMEM Installation Guide](https://docs.nvidia.com/nvshmem/release-notes-install-guide/install-guide/abstract.html).
+#### 2.2 Install GDRCopy and load the gdrdrv kernel module
```
youkaichao, line 53: since both gdrcopy and turning on IBGDA support requires doing something at the sudo level, i think supporting IBGDA directly via configuring the driver looks easier.

any benefit for the cpu-assist mode?

sphish: I have the same question. The performance of the cpu-assisted mode is significantly worse than the gpu mode, and it also introduces an extra dependency on gdrcopy. Why are we adding this mode back?

setup.py
```diff
@@ -29,26 +29,42 @@
...
     if int(os.getenv('DISABLE_SM90_FEATURES', 0)):
         # Prefer A100
+        print("Not using SM_90")
         os.environ['TORCH_CUDA_ARCH_LIST'] = os.getenv('TORCH_CUDA_ARCH_LIST', '8.0')
 
         # Disable some SM90 features: FP8, launch methods, and TMA
         cxx_flags.append('-DDISABLE_SM90_FEATURES')
         nvcc_flags.append('-DDISABLE_SM90_FEATURES')
 
+        # Add architecture flags to nvcc_dlink for the final linking step
```
sphish, line 44: This doesn't seem necessary here, as the Torch CUDAExtension will add gencode options automatically according to the `TORCH_CUDA_ARCH_LIST`.

setup.py
```diff
@@ -29,26 +29,42 @@
...
         include_dirs.extend([f'{nvshmem_dir}/include'])
         library_dirs.extend([f'{nvshmem_dir}/lib'])
-        nvcc_dlink.extend(['-dlink', f'-L{nvshmem_dir}/lib', '-lnvshmem'])
-        extra_link_args.extend(['-l:libnvshmem.a', '-l:nvshmem_bootstrap_uid.so', f'-Wl,-rpath,{nvshmem_dir}/lib'])
+        nvcc_dlink.extend(['-dlink', f'-L{nvshmem_dir}/lib', '-lnvshmem_device'])
+        extra_link_args.extend(['-l:libnvshmem_host.so', '-l:libnvshmem_device.a', f'-Wl,-rpath,{nvshmem_dir}/lib'])
 
     if int(os.getenv('DISABLE_SM90_FEATURES', 0)):
         # Prefer A100
+        print("Not using SM_90")
```
sphish, line 37: We have already printed TORCH_CUDA_ARCH_LIST below, so there's no need to print it here.

setup.py
```diff
@@ -29,26 +29,42 @@
         sources.extend(['csrc/kernels/internode.cu', 'csrc/kernels/internode_ll.cu'])
         include_dirs.extend([f'{nvshmem_dir}/include'])
         library_dirs.extend([f'{nvshmem_dir}/lib'])
-        nvcc_dlink.extend(['-dlink', f'-L{nvshmem_dir}/lib', '-lnvshmem'])
-        extra_link_args.extend(['-l:libnvshmem.a', '-l:nvshmem_bootstrap_uid.so', f'-Wl,-rpath,{nvshmem_dir}/lib'])
+        nvcc_dlink.extend(['-dlink', f'-L{nvshmem_dir}/lib', '-lnvshmem_device'])
+        extra_link_args.extend(['-l:libnvshmem_host.so', '-l:libnvshmem_device.a', f'-Wl,-rpath,{nvshmem_dir}/lib'])
```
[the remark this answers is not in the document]
seth-howell: I was validating this on my local system, and it seems that some NVIDIA wheels which only include C++ binaries (NVSHMEM, NCCL, etc.) don't have __init__.py files in them so it's impossible to do this right now (nvshmem.__file__ is None). I've raised a ticket internally to fix that.

third-party/README.md
sphish: I see. I think this also needs to be modified. https://github.com/deepseek-ai/DeepEP/blob/079c5a4f8b8529ae4c3b4ac3d50709230e5e13f1/deep_ep/buffer.py#L86

setup.py
sphish: I tested with https://developer.download.nvidia.com/compute/nvshmem/redist/libnvshmem/linux-x86_64/libnvshmem-linux-x86_64-3.3.9_cuda12-archive.tar.xz and did not encounter similar issues. Also, hardcoding sm_90 is not ideal, since, for example, SGLang has already started using DeepEP on sm_100.

setup.py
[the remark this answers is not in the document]
seth-howell: Done.

sphish:
LGTM, any suggestion? @youkaichao

youkaichao:
LGTM now, thanks!

ishandhanani:
~@youkaichao - we're having some trouble when running SGLang + DeepEP after this version bump. Specifically we see some `cuMemCreate failed` errors. Curious on if you've seen that recently after this PR?~

Resolved by fixing cuda graph bs in sglang

youkaichao:
> we're having some trouble when running SGLang + DeepEP after this version bump. Specifically we see some `cuMemCreate failed` errors. Curious on if you've seen that recently after this PR?

sorry i have no ideas

alpha-baby:
> @youkaichao - we're having some trouble when running SGLang + DeepEP after this version bump. Specifically we see some `cuMemCreate failed` errors. Curious on if you've seen that recently after this PR?-在此版本升级后，我们在运行 SGLang+DeepEP 时遇到了一些问题。具体来说，我们看到了一些#0 错误。想知道在这次公关之后，你最近是否看到了这一点？

can you show detail log?

HPC4AI:
> ~@youkaichao - we're having some trouble when running SGLang + DeepEP after this version bump. Specifically we see some `cuMemCreate failed` errors. Curious on if you've seen that recently after this PR?~
> 
> Resolved by fixing cuda graph bs in sglang

Hello, I’m also encountering the same issue. Could you share how you resolved it? Why does CUDA Graph cause this error?
Thanks

ishandhanani:
> > ~@youkaichao - we're having some trouble when running SGLang + DeepEP after this version bump. Specifically we see some `cuMemCreate failed` errors. Curious on if you've seen that recently after this PR?~
> > Resolved by fixing cuda graph bs in sglang
> 
> Hello, I’m also encountering the same issue. Could you share how you resolved it? Why does CUDA Graph cause this error? Thanks


NVSHMEM during init allocs memory. CUDA graph was taking too much mem causing the fail

PR #305, opened 2025-07-15
correct the wqe_idx in rdma write wqe

GuangguanWang:
correct the wqe_idx in rdma write wqe when num_wqes > 1 in nvshmemi_ibgda_put_nbi_warp.


sphish:
Thanks!

PR #313, opened 2025-07-18
Fix the address of dispatch_rdma_recv_count_buffer in LL Layout to avoid cleaning after each change in hidden_size/token_num.

yilin-void:
Each time the `hidden_size` or `num_max_dispatch_tokens_per_rank` in LL dispatch/combine changes, the address of `dispatch_rdma_recv_count_buffer` also changes. This causes the address pointed to by `next_clean` in the previous kernel to not be actually used, necessitating an additional clean operation (e.g., calling `clean_low_latency_buffer`). 

In this PR, I have placed the `dispatch_rdma_recv_count_buffer` that needs to be cleaned at the beginning of the entire RDMA buffer (aligned to 128B to ensure cacheline alignment). This ensures that regardless of changes in `hidden_size` or `num_max_dispatch_tokens_per_rank`, as long as `num_experts` remains the same, the correct `dispatch_rdma_recv_count_buffer` address will always be cleaned.


sphish:
Why is it necessary to modify `hidden_size` or `num_max_dispatch_tokens_per_rank` during runtime?

yilin-void:
> Why is it necessary to modify `hidden_size` or `num_max_dispatch_tokens_per_rank` during runtime?

For example, if we want to use different types of data for dispatch/combine, we can achieve this by packing these data into a bf16 tensor. However, the hidden size of the packed bf16 tensor will be inconsistent. Additionally, the maximum number of tokens that each rank will actually send will vary with the batch size and sequence length.

PR #317, opened 2025-07-21
Avoid hang due to small rdma_chunk_size

ZhiyiHu1999:
For example, if we have two nodes, `num_warps_per_forwarder` will be 16 / 2 = 8. 

https://github.com/deepseek-ai/DeepEP/blob/e6012370b01a86e821ace99193f5d912b3f66d06/csrc/kernels/internode.cu#L1550

If `num_max_rdma_chunked_send_tokens` is 4 < 8, then `token_end_idx` - `token_start_idx` = 4, the 'forwarder_nvl_head' for subwarp 4, 5, 6, 7 will always be 0 and `min_head` in forwarder kCoordinator will always be 0. So, the `NVLSender `warp won't work after the NVL buffer is full for the first time because the head value it reads is always 0.

https://github.com/deepseek-ai/DeepEP/blob/e6012370b01a86e821ace99193f5d912b3f66d06/csrc/kernels/internode.cu#L1570-L1605


LyricZhao:
Thanks!

PR #345, opened 2025-07-31
Support 10-bit LogFMT Combine

zheanxu:


fzyzcjy:
FYI when testing this PR, it seems to get stuck. I am not sure whether this is known issue or not, thus firstly report it instead of digging into it. If you need me to check and fix just ping me!

```
python test_low_latency.py --num-processes 4 --use-logfmt
```

without `--use-logfmt`: works well
with logfmt: stuck at

```
Allocating buffer size: 2146.961792 MB ...
/opt/nvshmem_src/src/host/transport/transport.cpp:nvshmemi_transport_init:275: init failed for transport: IBGDA
/opt/nvshmem_src/src/host/transport/transport.cpp:nvshmemi_transport_init:275: init failed for transport: IBGDA
/opt/nvshmem_src/src/host/transport/transport.cpp:nvshmemi_transport_init:275: init failed for transport: IBGDA
/opt/nvshmem_src/src/host/transport/transport.cpp:nvshmemi_transport_init:275: init failed for transport: IBGDA
```

LyricZhao (approved):

PR #349, opened 2025-07-31
fix combined RdmaReciver return slow head

a1372422617:
internode-combined {RDMAReiciver} need return expect_head + 1,but return the min_head of rdma_receiver_rdma_head,
although this smaller head update in the response won't cause a hang, nor lead to accuracy issues, it still represents a potential bug.

### DIFF
--- a/csrc/kernels/internode.cu
+++ b/csrc/kernels/internode.cu
@@ -1789,7 +1789,7 @@ combine(int4* combined_x, float* combined_topk_weights,
                     for (int i = 0; i < kNumRDMAReceivers; ++ i) if (not rdma_receiver_retired[i])
                         min_head = min(min_head, rdma_receiver_rdma_head[i][dst_rdma_rank]);
                     if (min_head != std::numeric_limits<int>::max() and min_head >= last_rdma_head + num_max_rdma_chunked_send_tokens and lane_id < kNumRDMARanks) {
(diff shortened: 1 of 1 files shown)


sphish:
https://github.com/deepseek-ai/DeepEP/blob/26cf250a3388c770ee628d2fd0271e1c6c52157a/csrc/kernels/internode.cu#L1731
I understand that you mean `rdma_receiver_rdma_head` should be updated to `expected_head + 1` here. However, the current warp has not yet waited for the token pointed to by `expected_head` to arrive. Updating `rdma_head` to `expected_head + 1` at this point might introduce bugs.
In contrast, `nvl_head` uses a different update approach: it updates the head to `expected_head + 1` after `combine_token`, that is, after the token pointed to by `expected_head` has been processed. If we want to update `rdma_head` more promptly, we should also update it to `expected_head + 1` after `combine_token`, just like `nvl_head` does.

https://github.com/deepseek-ai/DeepEP/blob/26cf250a3388c770ee628d2fd0271e1c6c52157a/csrc/kernels/internode.cu#L1671-L1681

We are considering removing the head mechanism in the next refactoring. Since this slow update issue only affects a very small number of cases (for example, when the `rdma_buffer_size` is set to a very small value), we will leave it as is for now.

a1372422617:
I realize this is a rather silly question, but I truly appreciate your response nonetheless

Issue #314, opened 2025-07-18
Internode Combine Timeout


ZhiyiHu1999:
Hello! I found under this configuration, the **internode combine** will timeout: (The variable names I use here are the same as the ones used in `test_internode.py`)

**num_sms = 48
nvl_chunk_size = 32
nvl_buffer_size = 80
rdma_chunk_size = 20
rdma_buffer_size = 128**

For my experiment setup: 
**Two nodes, each with 8 * H20 GPU and 4 * NIC, ROCE**

Could you try the config in your environment to confirm or deny this. Thanks a lot!

sphish:
I am able to reproduce this issue. It seems to occur when the `nvl_buffer_size` is set to a small value. I will look into it and investigate further.

kzlxd:
how about this problem？@sphish ，this problem  Stable reproduction during continuous stress testing

Skylion007:
Also can replicate unfortunately.

sphish:
@Skylion007 @kzlxd Did you also use configs similar to those mentioned in the issue?

PR #353, opened 2025-08-06
fix combine timeout due to delayed forwarder min head update

ZhiyiHu1999:
This PR provides a fix for #314.

The configuration that causes the combine timeout in #314 is:
num_sms = 48
nvl_chunk_size = 32
nvl_buffer_size = 80
rdma_chunk_size = 20
rdma_buffer_size = 128

In `kNVLAndRDMAForwarder`, we first read `expected_head`, then compare with the `tail`, combine the token, issue RDMA send, finally update `forwarder_nvl_head` with `expected_head + 1`, then the coordinator updates the `min_head`.

In the above configuration, a rare case will happen: in the **first** iteration of the below 'for' loop, some threads will be trapped in the while loop comparing `nvl_tail` and `nvl_head`. As we update the  `forwarder_nvl_head` at the end of the iteration, even if  `eapected_head` is larger than the `forwarder_nvl_head`, it won't get updated, same for `min_head`.

https://github.com/deepseek-ai/DeepEP/blob/26cf250a3388c770ee628d2fd0271e1c6c52157a/csrc/kernels/internode.cu#L1646-L1664

So, I just simply update `forwarder_nvl_head` with `expected_head` (not `expected_head + 1` in the fix) for the rare cases. 

Besides, for the `kRDMAReceiver`, the design is to update `rdma_receiver_rdma_head` with `expected_head`, before consuming the token. So maybe an update for `rdma_receiver_rdma_head` with `expected_head + 1` should be added after the token consuming in the last 'for' loop iteration. I guess there may be rare cases where this could potentially cause a timeout. But I haven't found a configuration for that.

### DIFF
--- a/csrc/kernels/internode.cu
+++ b/csrc/kernels/internode.cu
@@ -1647,14 +1647,15 @@ combine(int4* combined_x, float* combined_topk_weights,
                     // Read expected head
                     EP_STATIC_ASSERT(kNumRDMARanks <= 32, "Invalid number of RDMA peers");
                     int expected_head = -1;
-                    if (lane_id < NUM_MAX_NVL_PEERS)
+                    if (lane_id < NUM_MAX_NVL_PEERS) {
                         expected_head = ld_nc_global(combined_nvl_head + token_idx * NUM_MAX_NVL_PEERS + lane_id);
+                        expected_head < 0 ? (forwarder_nvl_head[warp_id][lane_id] = -expected_head - 1) : (forwarder_nvl_head[warp_id][lane_id] = expected_head);
+                    }
 
                     // Wait lanes to be ready
                     start_time = clock64();
                     while (cached_nvl_channel_tail_idx <= expected_head) {
                         cached_nvl_channel_tail_idx = ld_acquire_sys_global(nvl_channel_tail.buffer(lane_id));
-                        if (forwarder_nvl_head[warp_id][lane_id] < expected_head) forwarder_nvl_head[warp_id][lane_id] = expected_head;
 
                         // Timeout check
                         if (clock64() - start_time > NUM_TIMEOUT_CYCLES and lane_id < NUM_MAX_NVL_PEERS) {
@@ -1852,6 +1853,7 @@ void combine(cudaDataType_t type,
     EP_HOST_ASSERT(num_forwarder_warps > NUM_MAX_NVL_PEERS and num_forwarder_warps % num_rdma_ranks == 0);
     EP_HOST_ASSERT(num_max_nvl_chunked_recv_tokens % num_rdma_ranks == 0);
     EP_HOST_ASSERT(num_max_nvl_chunked_recv_tokens / num_rdma_ranks > std::max(num_max_rdma_chunked_send_tokens, num_max_nvl_chunked_send_tokens));
+    EP_HOST_ASSERT(num_max_nvl_chunked_recv_tokens / num_rdma_ranks - num_warps_per_forwarder >= num_max_nvl_chunked_send_tokens);
     EP_HOST_ASSERT(num_max_rdma_chunked_send_tokens >= num_warps_per_forwarder);
     EP_HOST_ASSERT(type == CUDA_R_16BF);
 
--- a/csrc/kernels/internode.cu
+++ b/csrc/kernels/internode.cu
@@ -1654,6 +1654,7 @@ combine(int4* combined_x, float* combined_topk_weights,
                     start_time = clock64();
                     while (cached_nvl_channel_tail_idx <= expected_head) {
                         cached_nvl_channel_tail_idx = ld_acquire_sys_global(nvl_channel_tail.buffer(lane_id));
+                        if (forwarder_nvl_head[warp_id][lane_id] < expected_head) forwarder_nvl_head[warp_id][lane_id] = expected_head;
 
                         // Timeout check
                         if (clock64() - start_time > NUM_TIMEOUT_CYCLES and lane_id < NUM_MAX_NVL_PEERS) {


sphish:
Thank you for your submission!
I tested your changes using a similar configuration, but on my test env, timeouts still occasionally occur.

I have recently looked into the issue mentioned in #314 and have made some findings. The head update issue you mentioned is one of the factors, and there is also some related discussion in #349.

Another factor is that each forwarder warp group performs a `sync_large_warp` after each round of combining tokens before proceeding to the next round and updating the head. This also results in the head not being updated promptly.https://github.com/deepseek-ai/DeepEP/blob/510719682c3d89ab51e25318ee378e793bce0e2f/csrc/kernels/internode.cu#L1683

To fully resolve this issue, I think we need to:

1. Update the head both before and after `combine_token`
2. Assert that `nvl_buffer_size_per_rdma_rank - num_warps_per_forwarder >= nvl_chunk_size`

I have done some preliminary testing of this approach, but I have not yet found a way to prove its correctness.

kzlxd:
Update the head both before and after combine_token ? is this necessary？@sphish @ZhiyiHu1999

ZhiyiHu1999:
> Update the head both before and after combine_token ? is this necessary？@sphish @ZhiyiHu1999

As @sphish explained, I think this makes sense and have committed a new modification for it.

kzlxd:
As I understand it, updating the head before and after a warp should not cause a hang. @ZhiyiHu1999

kzlxd:
Just put it in front, like the receiver does.

MARD1NO:
Do you encounter combine timeout when using default config?

Skylion007:
This may not be perfect, but it seems like this does at least make the newer DeepEP usable on my setup. Any progress on getting this merged @kzlxd?

sphish:
@Skylion007 Did your setup use a similar configuration, with a large RDMA chunk size but a relatively small RDMA buffer? I had assumed that most people would not use such a configuration.

Skylion007:
I am just using the default settings but am encountering the issue very reliably on H200s. Reverting to an older DeepEP commit from March seems to solve the issue though. This PR still times out unfortunately, but it way less frequent than the previous commit. No idea why it's such a reliable deadlock on H200s internode or why this PR seems to improve the issue so much though. I am unable to use a recent commit without this patch (even though it still timeouts/deadlocks sometimes).

MARD1NO:
> I am just using the default settings but am encountering the issue very reliably on H200s. Reverting to an older DeepEP commit from March seems to solve the issue though. This PR still times out unfortunately, but it way less frequent than the previous commit. No idea why it's such a reliable deadlock on H200s internode or why this PR seems to improve the issue so much though. I am unable to use a recent commit without this patch (even though it still timeouts/deadlocks sometimes).

Hi, which commit do you use?... I also encounter timeout occasionally in H200s ....

my setting is: 
```
DEEP_EP_DISPATCH_NVL_CHUNK_SIZE=44
DEEP_EP_DISPATCH_RDMA_CHUNK_SIZE=12
DEEP_EP_COMBINE_NVL_CHUNK_SIZE=3
DEEP_EP_COMBINE_RDMA_CHUNK_SIZE=16
```

sphish:
@Skylion007 Before any further progress is made on this issue, you can try increasing the RDMA buffer size or NVL buffer size. This may help alleviate your problem.

Skylion007:
a84a248 is the commit that works better for me on H200

MARD1NO:
> [a84a248](https://github.com/deepseek-ai/DeepEP/commit/a84a24808fb0ea732f49b874cc456a69dde69076) is the commit that works better for me on H200

Have you tried this commit bdd119f8b249953cab366f4d737ad39d4246fd7e ? It seems work well...

Skylion007:
I am using the default configs from https://github.com/deepseek-ai/DeepEP/blob/e3908bf5bd0cc6265bcb225d15cd8c996d4759ef/deep_ep/buffer.py#L206 and I am still getting deadlock on my H200 machines without this patch @sphish should I still increase the defaults for chunk_size.

Here are my best tuned configs, but I am not using them yet:
```
[tuning] Best dispatch (FP8): SMs 24, NVL chunk 40, RDMA chunk 32, transmit: 24736.00 us, notify: 113.35 us, BW: 4.42 GB/s (RDMA), 8.83 GB/s (NVL)
[tuning] Best dispatch (FP8): SMs 24, NVL chunk 16, RDMA chunk 32, transmit: 24699.00 us, notify: 161.50 us, BW: 4.45 GB/s (RDMA), 8.97 GB/s (NVL)
[tuning] Best dispatch (FP8): SMs 24, NVL chunk 16, RDMA chunk 32, transmit: 24718.00 us, notify: 144.36 us, BW: 4.42 GB/s (RDMA), 8.88 GB/s (NVL)
[tuning] Best dispatch (FP8): SMs 24, NVL chunk 16, RDMA chunk 32, transmit: 24690.00 us, notify: 141.19 us, BW: 4.43 GB/s (RDMA), 8.93 GB/s (NVL)
[tuning] Best dispatch (BF16): SMs 24, NVL chunk 4, RDMA chunk 12, transmit: 47846.00 us, notify: 119.53 us, BW: 4.43 GB/s (RDMA), 8.85 GB/s (NVL)
[tuning] Best dispatch (BF16): SMs 24, NVL chunk 4, RDMA chunk 32, transmit: 47720.00 us, notify: 286.33 us, BW: 4.44 GB/s (RDMA), 8.92 GB/s (NVL)
[tuning] Best dispatch (BF16): SMs 24, NVL chunk 36, RDMA chunk 24, transmit: 47772.00 us, notify: 235.92 us, BW: 4.46 GB/s (RDMA), 8.99 GB/s (NVL)
[tuning] Best dispatch (BF16): SMs 24, NVL chunk 4, RDMA chunk 32, transmit: 47730.00 us, notify: 283.09 us, BW: 4.44 GB/s (RDMA), 8.96 GB/s (NVL)
[tuning] Best combine: SMs 24, NVL chunk 2, RDMA chunk 12, transmit: 47718.00 us, notify: 130.50 us, BW: 4.44 GB/s (RDMA), 8.88 GB/s (NVL)
[tuning] Best combine: SMs 24, NVL chunk 2, RDMA chunk 12, transmit: 47677.00 us, notify: 172.74 us, BW: 4.45 GB/s (RDMA), 8.97 GB/s (NVL)
[tuning] Best combine: SMs 24, NVL chunk 2, RDMA chunk 12, transmit: 47693.00 us, notify: 152.15 us, BW: 4.44 GB/s (RDMA), 8.92 GB/s (NVL)
[tuning] Best combine: SMs 24, NVL chunk 2, RDMA chunk 12, transmit: 47702.00 us, notify: 148.73 us, BW: 4.47 GB/s (RDMA), 9.00 GB/s (NVL)
```

Skylion007:
Also using buffer_size from defaults in DeepEP via: get_nvl_buffer_size_hint, get_rdma_buffer_size_hint, etc

Skylion007:
I am also using megatron which sets the number of sms at 20. Not sure if that is part of the issue.

PR #357, opened 2025-08-09
delete some redundant logic

zhoutianzi666:


alpha-baby:
maybe reduce performance when directly use `cg::this_grid().sync();`

sphish:
The original code logic was intended to allow the NIC to start working as early as possible, thereby reducing latency. This is especially beneficial when the batch size is small. Have you tested the performance impact of your changes?

zhoutianzi666:
> The original code logic was intended to allow the NIC to start working as early as possible, thereby reducing latency. This is especially beneficial when the batch size is small. Have you tested the performance impact of your changes?

I tested, it does not hurt performance.

PR #371, opened 2025-08-21
fix potential hang in pollcq

Thunderbrook:
Hi, DeepEP will now execute pollcq before post wqe. The first time pollcq is performed, cqe is invalid (DeepEP does not check it now), but the existing logic will determine wqe_counter, when nvshmem is initialized, it will be initialized to 0xff. Therefore, when the queue depth is small, it may hang here. 

DeepEP:
```cpp
do {
        wqe_counter = HtoBE16(ld_na_relaxed(&cqe64->wqe_counter));
    } while ((static_cast<uint16_t>(static_cast<uint16_t>(idx) - wqe_counter - static_cast<uint16_t>(2)) < ncqes));
```

nvshmem:
```cpp
status = cudaMemset(cq_mobject->base.gpu_ptr, 0xff, cq_mobject->base.size);
```


kzlxd:
There is a uint16_t conversion here, so it shouldn't get stuck, right? @sphish @Thunderbrook

PR #376, opened 2025-08-25
Tiny fix comments about package format

fzyzcjy:
Realize this issue when checking message format as well as reading https://github.com/deepseek-ai/DeepEP/pull/341


shifangx:
Yes, Maybe we should change the comment as following
 `// Message package: index at source(int), 3 reserved int fields, hidden data, FP8 scales.`

PR #381, opened 2025-08-28
Fix duplicate calculation of num_tokens_per_rank

tianchongchong:
Each thread is responsible for one target rank. Replace each thread repeat the calculation kNumRanks times


LyricZhao:
Thanks!

PR #408, opened 2025-09-16
Support arbitrary hidden sizes in intranode combine kernel

varun-sundar-rabindranath:
Intranode combine kernels fail for arbitrary hidden size. 
The solution is to add a warpmask for __syncwarp.

Test:
 Tested with hidden-size 2880 (gpt-oss hidden-dim)


varun-sundar-rabindranath:
Requesting review from @LyricZhao as this PR touches `elect_one_sync` . Thanks 🙌

csrc/kernels/intranode.cu
```diff
@@ -837,24 +837,26 @@ combine(dtype_t* recv_x, float* recv_topk_weights,
                         out_dtypes[j] = static_cast<dtype_t>(values[j]);
 
 #ifndef DISABLE_SM90_FEATURES
+		    auto const warp_mask = __activemask();
```
LyricZhao, line 840: Sorry, `__activemask` can not ensure the mask is right. For example, even the hidden size makes all 32 lanes work together, but some lanes may wait for global memory load arrive and lead to a divergence.

LyricZhao (changes requested):

csrc/kernels/intranode.cu
varun-sundar-rabindranath: I see. Thanks @LyricZhao. I have updated the code to compute the active mask in the code directly. PTAL!

sphish:
https://github.com/deepseek-ai/DeepEP/pull/413 We have fixed this issue using an alternative approach, which also resolved the FP8 tests. However, the current internode kernels still do not support cases where `hidden_size % 128 != 0`. We plan to address this in the future refactoring.

Issue #358, opened 2025-08-11
[Proposal] Integrate some enhancements from TensorRT-LLM


yuantailing:
Hi, DeepEP team,

I'm a collaborator to the TensorRT-LLM project. First, thank you for creating such a performant library. We have successfully integrated it into our project. To better align with TensorRT-LLM's data format, we've made a few modifications to DeepEP. The specific commit we are using is [edf3ea2](https://github.com/deepseek-ai/DeepEP/commits/edf3ea2b086a393d3163bf2773eab69d9191cc01/). The main changes are as follows:

(1) Besides `torch.distributed.ProcessGroup`, support for initializing buffers with `mpi4py.MPI.Comm` ( @yuantailing )
(2) Change the dtype of `topk_idx` from `int64_t` to `int` and add an offset to expert id  ( @yifeizhang-c )
(3) Avoid cleaning after each change in hidden_size/token_num (Already merged in #313, @yilin-void )
(4) Add an FP4 low-latency dispatch kernel ( @yilin-void )

**Before we create pull requests, we would like to discuss whether the community would be open to these changes and how we can make them more general, especially for (2) and (4).**

For point (2), our proposal is to add a conditional type definition, such as `using topk_idx_t = int64_t` or `int` which depends on a macro.
For point (4), the FP4 dispatch is currently a separate kernel, but we could merge it into the original dispatch kernel by adding an input tensor to indicate that the scale is quantized outside of the kernel.

We hope these contributions will broaden DeepEP's applications and allow us to stay aligned with the main branch of your library.

Thanks
Tailing

sphish:
Thank you for your proposal! 

For (1) and (2), we are happy to merge these changes. 

Regarding the FP4 kernels, we haven’t yet found a way to use FP4 for data transfer without losing model performance, so we’re still unsure about the best approach moving forward. However, I think we can maintain an FP4 branch in the DeepEP repo to facilitate further exploration around FP4. If it's convenient, feel free to add me on WeChat (ID: Sphizzz) for further discussion.

yuantailing:
Added
(6) Support CUDA Graph for internode dispatch normal kernel (@yifeizhang-c )

0z5a:
Hi, I’d be happy to take a narrow follow-up for the MPI initialization path on hybrid-ep. @yuantailing @sphish 

I already have #730 on main, where I worked on communicator-aware topology/preflight handling and multi-rank failure propagation.

PR #422, opened 2025-09-23
Make dtype of topk_idx configurable

yuantailing:
This is the 2nd part of #358


csrc/kernels/configs.cuh
```diff
@@ -58,6 +58,20 @@ typedef int __nv_fp8x4_e4m3;
 typedef uint8_t __nv_fp8_storage_t;
 #endif
 
+namespace deep_ep {
+
+#ifndef TOPK_IDX_BITS
+#define TOPK_IDX_BITS 64
+#endif
+
+#define INT_BITS_T2(bits) int##bits##_t
```
sphish, line 67: Why not just define `INT_BITS_T` like this?

yuantailing: It's a trick of macros.
```cpp
#define TOPK_IDX_BITS 64
#define INT_BITS_T(bits) int##bits##_t
typedef INT_BITS_T(TOPK_IDX_BITS) topk_idx_t;
topk_idx_t topk_idx;
```
I got `error: ‘intTOPK_IDX_BITS_t’ does not name a type`

sphish:
Thanks!

PR #439, opened 2025-09-30
Mcore dev

Autumn1998:
1. Refine Document 
2. Add permute/unpermute to hide the D2D copy
3. Add support for the sync-free mode


SeanLi-OI:
Hi, @Autumn1998, I found that in
[hybrid_ep_buffer.py#L410](https://github.com/deepseek-ai/DeepEP/blob/f4f2c10aa57d3155a1320b0c26a6e70d5746d204/deep_ep/hybrid_ep_buffer.py#L410)

if use_host_meta=True，it just move the d2h sync from outside into deepep inside.

So if there is always a cuda sync and cannot avoid if I train a dropless MoE? Or is there a chance using a pin_memory buffer, and copy into it with non_blocking=True.

PR #459, opened 2025-10-16
Add guards for scales when the type is bf16

bnellnm:
I'm working on integrating the DeepEP hybrid kernels with vLLM and ran into problems when running with bf16 tokens.  I added some checks to allow for null scales for bf16 and it fixed the crashes I was running into.

cc @jershi425 , @Autumn1998

### DIFF
--- a/csrc/hybrid_ep/backend/hybrid_ep_backend.cuh
+++ b/csrc/hybrid_ep/backend/hybrid_ep_backend.cuh
@@ -204,15 +204,15 @@ struct dispatch_kernel_param_t{
   // Input buffers. These buffers are local buffers.
   const TOKEN_DATA_TYPE* attn_input_token;
   const float* attn_input_prob; // Needed by expert layer, so only valid in forward dispatch.
-  const float* attn_input_token_scaling_factor; // If input token is FP8 dtype, we need scaling factor for tokens.
+  const float* attn_input_token_scaling_factor = nullptr; // If input token is FP8 dtype, we need scaling factor for tokens.
   // Output buffers. These buffers are both local and remote buffers.
   TOKEN_DATA_TYPE* expert_output_token[MAX_NUM_OF_RANKS_PER_NODE];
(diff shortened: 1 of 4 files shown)


Autumn1998:
Hello, @bnellnm , in the early version, there was a bug when trying to open handles of scales in BF16 mode which caused crashes, but it should be fixed in https://github.com/deepseek-ai/DeepEP/pull/448.  Have you encountered this problem in the latest version? 
Otherwise, I understand that abandoning the scale on BF16 can also save some memory, but now the buffer will be reallocated if the config requires larger memory then pre-allocated buffer. Retaining the scale buffer allows us to avoid processing scales when continuously receiving inputs with different data formats (FP, BF16).

bnellnm:
> Hello, @bnellnm , in the early version, there was a bug when trying to open handles of scales in BF16 mode which caused crashes, but it should be fixed in #448. Have you encountered this problem in the latest version? Otherwise, I understand that abandoning the scale on BF16 can also save some memory, but now the buffer will be reallocated if the config requires larger memory then pre-allocated buffer. Retaining the scale buffer allows us to avoid processing scales when continuously receiving inputs with different data formats (FP, BF16).

Hi @Autumn1998 , I'll try again with the newer commit and let you know.

bnellnm:
> > Hello, @bnellnm , in the early version, there was a bug when trying to open handles of scales in BF16 mode which caused crashes, but it should be fixed in #448. Have you encountered this problem in the latest version? Otherwise, I understand that abandoning the scale on BF16 can also save some memory, but now the buffer will be reallocated if the config requires larger memory then pre-allocated buffer. Retaining the scale buffer allows us to avoid processing scales when continuously receiving inputs with different data formats (FP, BF16).
> 
> Hi @Autumn1998 , I'll try again with the newer commit and let you know.

@Autumn1998 I've just updated to the latest version of the branch and bf16 seems to be working fine w/o my changes.  I'll close this PR.  Thanks!

Issue #336, opened 2025-07-29
When testing with test_internode.py, the dispatch and combine bandwidths are far below expectations.


KrisLu999:
conclusion: After analysis, I believe DeepEP’s high‑throughput kernel is incompatible with the NVSHMEM environment variable NVSHMEM_ENABLE_NIC_PE_MAPPING, causing NVSHMEM to select the same NIC during initialization and thus resulting in overall low test bandwidth.
Environment information: 2 H800 servers, each equipped with 4 IB network cards(CX7).
test command：
node1: 
export NVSHMEM_IB_ENABLE_IBGDA=1
export MASTER_ADDR=192.168.0.157
export WORLD_SIZE=2
export RANK=0
export NCCL_SOCKET_IFNAME=bond0
export NVSHMEM_ENABLE_NIC_PE_MAPPING=1
export NCCL_IB_HCA=mlx5_0,mlx5_1,mlx5_6,mlx5_7
export NVSHMEM_HCA_PE_MAPPING="mlx5_0:1:2,mlx5_1:1:2,mlx5_6:1:2,mlx5_7:1:2"
python3 test_internode.py
node2:
export NVSHMEM_IB_ENABLE_IBGDA=1
export MASTER_ADDR=192.168.0.157
export WORLD_SIZE=2
export RANK=1
export NCCL_SOCKET_IFNAME=bond0
export NVSHMEM_ENABLE_NIC_PE_MAPPING=1
export NCCL_IB_HCA=mlx5_0,mlx5_5,mlx5_6,mlx5_9
export NVSHMEM_HCA_PE_MAPPING="mlx5_0:1:2,mlx5_1:1:2,mlx5_6:1:2,mlx5_7:1:2"
python3 test_internode.py
symptoms：
<img width="953" height="705" alt="Image" src="https://github.com/user-attachments/assets/71fd5d4c-8195-4e3e-8e65-57bbf5b0ad84" />
log（high Throughput）：
node1:
After inspecting the logs, I found that during NVSHMEM initialization all eight ranks only selected the first NIC to establish connections.
<img width="971" height="437" alt="Image" src="https://github.com/user-attachments/assets/12ad02fc-6533-46c0-81da-f5232bd97744" />
IB NIC traffic monitoring：
node1：
mlx5_0(ibp25s0):
<img width="849" height="705" alt="Image" src="https://github.com/user-attachments/assets/e026a5fe-81cd-4da9-8c24-94916c8cd5c7" />
[...]

PR #466, opened 2025-10-24
Allow NVSHMEM PE to NIC mapping to be initialized by DeepEP rank

smarterclayton:
The `nvshmemi_get_devices_by_distance` default initialization method in NVSHMEM does not work optimally for GPU configurations where 2 GPU and 2 RDMA NIC share a PCIe bus, such as the x86 based GCP A3 Ultra H200 and A4 B200 instance types: https://cloud.google.com/compute/docs/gpus/gpu-network-bandwidth#h200-gpus. GPU0 and GPU1 (on two independent processes) observe NIC0 and NIC1 on the same PCIe switch as being equidistant and the default configuration for DeepEP + NVSHMEM results in both GPUs being assigned NIC0 from `nvshmemi_get_devices_by_distance`, halving the observed bandwidth for RDMA in test_internode.py and in vLLM wide-EP (because 4 of 8 NIC are enabled).

The alternative is a static mapping between GPU host index (PE) and NIC index (HCA), but the `NVSHMEMX_INIT_WITH_UNIQUEID` initialization method bypasses setting `mype_node` and `npes_node`.

The `nvshmemi_boot_handle.pg_rank` for this initialization method is always 0 and the `nvshmem_boot_handle.pg_size` is always 2, and as a result mype_node and npes_node are set to 0 and 2 respectively for all PEs initializing. This prevents `NVSHMEM_ENABLE_NIC_PE_MAPPING=1` from selecting from a static list of devices by mype_node / local rank in `transport.cpp#nvshmemi_setup_connections`:

    selected_devices[0] =
      nvshmemi_state->mype_node % (tcurr->n_devices > 0
        ? tcurr->n_devices : 1);

To allow static assignment, we introduce a new `DEEP_EP_DEVICE_TO_HCA_MAPPING` environment variable during `deep_ep.Buffer` initialization that accepts `<cuda_device_id>:<HCA_name>:<HCA_port>` and uses `torch.cuda.current_device()` to set `NVSHMEM_HCA_LIST` to the appropriate `<HCA_name>:<HCA_port>`, or error if no such device was listed.

We are proposing the change to DeepEP because the choice of initialization method determines how HCA to GPU binding can be achieved across multiple existing NVSHMEM versions.  Because vLLM 0.11.0 uses CUDA_VISIBLE_DEVICES to associate GPU devices to each rank, we reverse map the visible devices to the current cuda device.

On GCP we would propose this configuration for all full host workloads on A3U and A4 instance types (H200/B200 + x86 with shared PCIe hubs per 2 NIC / 2 GPU):

```
DEEP_EP_DEVICE_TO_HCA_MAPPING=0:mlx5_0:1,1:mlx5_1:1,2:mlx5_2:1,3:mlx5_3:1,4:mlx5_4:1,5:mlx5_5:1,6:mlx5_6:1,7:mlx5_7:1
```

Tested with NVSHMEM 3.3.20 and vLLM 0.11.0 and vLLM main @ 2025/10/23.  

Co-Authored-By: Keon Jang <keonjang@google.com>

### DIFF
--- a/deep_ep/buffer.py
+++ b/deep_ep/buffer.py
@@ -103,6 +103,8 @@ class Buffer:
         # Synchronize NVSHMEM unique IDs
         root_unique_id = None
         if self.runtime.get_num_rdma_ranks() > 1 or low_latency_mode:
+            self._setup_device_hca_mapping()
+
             # Enable IBGDA
             assert num_qps_per_rank > 0
             os.environ['NVSHMEM_DISABLE_P2P'] = '0' if allow_nvlink_for_low_latency_mode else '1'
@@ -134,6 +136,29 @@ class Buffer:
         # Make CPP runtime available
         self.runtime.sync(device_ids, ipc_handles, root_unique_id)
         assert self.runtime.is_available()
+
+    def _setup_device_hca_mapping(self):
+        """
+        Set up device to NIC mapping using DEEP_EP_DEVICE_TO_HCA_MAPPING environment variable.
+        The mapping format is: "0:mlx5_0:1,1:mlx5_1:1,..." where each entry maps a CUDA device ID
+        to an HCA name separated by colon. HCA name can include additional suffixes like ":1".
+        """
+        if 'DEEP_EP_DEVICE_TO_HCA_MAPPING' in os.environ:
+            device_mapping = {}
+            mapping_str = os.environ['DEEP_EP_DEVICE_TO_HCA_MAPPING']
+            # Parse mapping string like "0:mlx5_0:1,1:mlx5_1:1,..."
+            for mapping in mapping_str.split(','):
+                assert ':' in mapping, f"Invalid mapping format '{mapping}' in DEEP_EP_DEVICE_TO_HCA_MAPPING. Expected format: '<device_id>:<hca_name>'"
+                parts = mapping.split(':', 1)  # Split only on first colon
+                device_id = int(parts[0])
+                hca_name = parts[1]  # Keep the rest as HCA name (including :1)
+                device_mapping[device_id] = hca_name
+
+            # Get current device and set appropriate HCA
+            current_device = torch.cuda.current_device()
+            assert current_device in device_mapping, f"Current CUDA device {current_device} not found in DEEP_EP_DEVICE_TO_HCA_MAPPING"
+            os.environ['NVSHMEM_ENABLE_NIC_PE_MAPPING'] = '1'
+            os.environ['NVSHMEM_HCA_LIST'] = device_mapping[current_device]
 
     def destroy(self):
         """
--- a/deep_ep/buffer.py
+++ b/deep_ep/buffer.py
@@ -156,6 +156,13 @@ class Buffer:
 
             # Get current device and set appropriate HCA
             current_device = torch.cuda.current_device()
+            # Translate CUDA_VISIBLE_DEVICES
+            if 'CUDA_VISIBLE_DEVICES' in os.environ:
+                visible_devices = os.environ['CUDA_VISIBLE_DEVICES'].split(",")
+                assert len(visible_devices) > current_device, f"CUDA_VISIBLE_DEVICES has {len(visible_devices)} entries which is fewer than the current device {current_device}"
+                assert visible_devices[current_device].isdigit(), f"DEEP_EP_DEVICE_TO_HCA_MAPPING requires CUDA_VISIBLE_DEVICES to contain integer indices"
+                current_device = int(visible_devices[current_device])
+
             assert current_device in device_mapping, f"Current CUDA device {current_device} not found in DEEP_EP_DEVICE_TO_HCA_MAPPING"
             os.environ['NVSHMEM_ENABLE_NIC_PE_MAPPING'] = '1'
             os.environ['NVSHMEM_HCA_LIST'] = device_mapping[current_device]


sphish:
Why not just set the NVSHMEM environment variables directly?

smarterclayton:
They have to be set per rank, so we would need to change all callers of deep_ep.Buffer to set those environment variables before invoking, including `test_low_latency` and `test_internode` (which we use to validate network performance of the supporting RDMA clusters).  Since we expected all consumers of DeepEP to need this mapping on our H200 and B200 VM families, we wanted to push the fix as low in the software stack as possible (so below SGLang, vLLM, NeMO, etc) so that those frameworks don't have to change how they call DeepEP.  We considered changing NVSHMEM but that would likely still require a change to Buffer initialization to pass local rank.

sphish:
NVSHMEM stores local rank information, for example in `mype_node` in the code.  
I believe the issue you’re experiencing is quite common — applications using NVSHMEM outside of DeepEP are likely to encounter it as well. It would be best to submit a PR to NVSHMEM to resolve this problem.

sphish:
@seth-howell Could you help have a look at this？

keonjang-google:
`mype_node` is 0 for all the PEs because DeepEP initializes NVSHMEM separately for each rank IIUC.

seth-howell:
@sphish
When working with multiple instances of NVSHMEM, there's no way for us to know which NIC is claimed by a process in the separate instance. The best way to do this currently is to set the NVSHMEM mapping environment variables.
> Why not just set the NVSHMEM environment variables directly?
It looks like the main benefit of using a DeepEP scoped environment variable is that you only have to set one, rather than both.

> I believe the issue you’re experiencing is quite common — applications using NVSHMEM outside of DeepEP are likely to encounter it as well.

The topology detection only fails when both cases are true:
1. The application launches multiple instances of NVSHMEM within a single application.
2. There are two GPUs and two NICs at equal distance from each other in the PCIe hierarchy.
Any applications not satisfying both conditions will run fine without any environment variables set.

The google use-case of DeepEP internode + the above hardware configurations is the first time we have seen this problem.

@smarterclayton 
> They have to be set per rank, so we would need to change all callers of deep_ep.Buffer to set those environment variables before invoking, **including test_low_latency**

The low_latency kernels only create one instance of NVSHMEM, so the topology detection and assignment work correctly in that case, properly distributing one unique NIC per process. Please let me know if you've had a different experience.


This solution is correct and likely the easiest to implement. FWIW, I would recommend merging it to unblock Google.

*Other possible solutions*

_DeepEP internode specific - requires changes in DeepEP_
If the internode initialization code is modified to instantiate NVSHMEM once across all processes and use NVSHMEMX_TEAM_SAME_MYPE_NODE to determine when to send over RDMA, it will assign the NICs properly.
This is what the low_latency kernel is doing today.

_More general, but potentially limited by environment - No required changes in DeepEP_
I believe it was @keonjang-google who suggested we look into adding GPU index as a consideration during topology detection. This is valid in some environments, but may not work in others.
An applications launched with Slurms gres features, or any application relying on CUDA_VISIBLE_DEVICES on a per-process basis would enumerate a subset of devices in an arbitrary order to the application. At the very least, this would require querying PCIe BDFs of GPUs not used in the current NVSHMEM instance and assigning them an ID independent from the CUDA enumeration. This would take some time to implement.
There are other edge cases (kata containers, VMs) where the GPUs may not even be visible at the PCIe level.

vimalKeshu:
This commit may potentially address this issue, https://github.com/deepseek-ai/DeepEP/issues/336, as well.
cc: @KrisLu999, @klauseshen

gangxie112:
Why not use the way of used by NCCL, providing the topo?

vimalKeshu:
@smarterclayton 

I would like to share my finding regarding below statement mentioned in your PR description:

> The nvshmemi_boot_handle.pg_rank for this initialization method is always 0 and the nvshmem_boot_handle.pg_size is always 2, and as a result mype_node and npes_node are set to 0 and 2 respectively for all PEs initializing.

Actually, nvshmemi_boot_handle.pg_rank and nvshmem_boot_handle.pg_size are set by RDMA_RANK and RDMA_WORLD_SIZE in normal inter node communication mode from deepEP code.

Code reference:

DeepEP code flow to setup NVSHMEM attribute: 

1. https://github.com/deepseek-ai/DeepEP/blob/main/csrc/deep_ep.cpp#L362-L364
2. https://github.com/deepseek-ai/DeepEP/blob/main/csrc/kernels/runtime.cu#L53

Code from NVSHMEM source, init.cu file
```cpp
int nvshmemx_set_attr_uniqueid_args(const int myrank, const int nranks,
                                    const nvshmemx_uniqueid_t *uid,
                                    nvshmemx_init_attr_t *nvshmem_attr) {
    assert(nvshmem_attr != NULL);
    nvshmemx_init_args_t *init_args = (nvshmemx_init_args_t *)(&(nvshmem_attr->args));
    nvshmemx_uniqueid_args_t *uid_args = &((init_args)->uid_args);

    /* Save to uid_args */
    uid_args->id = const_cast<nvshmemx_uniqueid_t *>(uid);  // Don't deepcopy as we are saving a ptr
    uid_args->myrank = myrank;
    uid_args->nranks = nranks;
    return (0);
}
```
What does it mean, in case of 4 Nodes: 

RDMA_RANK: [0, 1, 2, 3] and RDMA_WORLD_SIZE = 4
For Node:0, ranks (0 to 7):     nvshmemi_boot_handle.pg_rank=0 and nvshmem_boot_handle.pg_size=4
For Node:1, ranks (8 to 15):    nvshmemi_boot_handle.pg_rank=1 and nvshmem_boot_handle.pg_size=4
For Node:2, ranks (16 to 23):  nvshmemi_boot_handle.pg_rank=2 and nvshmem_boot_handle.pg_size=4
For Node:3, ranks (24 to 31):  nvshmemi_boot_handle.pg_rank=3 and nvshmem_boot_handle.pg_size=4



----------------------------------------------------------------------------

Regarding below statement from your PR description:

> The alternative is a static mapping between GPU host index (PE) and NIC index (HCA), but the NVSHMEMX_INIT_WITH_UNIQUEID initialization method bypasses setting mype_node and npes_node.

I have intuition that mype_node and npes_node are also derived from RDMA_RANK and RDMA_WORLD_SIZE, somewhere in NVSHMEM code while initializing the nvshmem state. However, I couldn't find the concrete code evidence which can support my intuition.

----------------------------------------------------------------------------

I see one drawback of your patch in my understanding (I hope, I am wrong !).

Your patch can make NVSHMEM_ENABLE_NIC_PE_MAPPING environment variable irrelevant ! If I have understood your patch correctly, you are mapping single NIC to device by passing environment variable per process

```python
os.environ['NVSHMEM_HCA_LIST'] = device_mapping[current_device]
```
I think because of this, each process may only see single choice in NVSHMEM_HCA_LIST ! 
Example, NVSHMEM_HCA_LIST=mlx5_0:1 for device:0 

It means, in NVSHMEM source code, it only sees single device.

NVSHMEM source code:
```cpp
int nvshmemi_setup_connections(nvshmemi_state_t *state) {
    int status = 0;
    nvshmem_transport_t *transports = (nvshmem_transport_t *)state->transports;
    nvshmem_transport_t tcurr;

    for (int i = 0; i < state->num_initialized_transports; i++) {
        if (!((state->transport_bitmap) & (1 << i))) continue;
        tcurr = transports[i];

        if (!(tcurr->attr & NVSHMEM_TRANSPORT_ATTR_CONNECTED)) {
            continue;
        }

        int devices_temp = tcurr->n_devices / state->npes_node;
        if (devices_temp == 0) devices_temp = 1;
        const int max_devices_per_pe = devices_temp;
        int selected_devices[max_devices_per_pe];
        int found_devices = 0;

        for (int j = 0; j < max_devices_per_pe; j++) {
            selected_devices[j] = -1;
        }

        // assumes symmetry of transport list at all PEs
        if (tcurr->n_devices <= 1) {
            /* return the index of the first available device.
             * -1 if no devices found.
             */
            selected_devices[0] = tcurr->n_devices - 1;
            found_devices++;
        } else if (nvshmemi_options.ENABLE_NIC_PE_MAPPING) {
            selected_devices[0] =
                nvshmemi_state->mype_node % (tcurr->n_devices > 0 ? tcurr->n_devices : 1);
            INFO(NVSHMEM_INIT, "NVSHMEM_ENABLE_NIC_PE_MAPPING = 1, setting dev_id = %d",
                 selected_devices[0]);
            found_devices++;
        } else {
            nvshmemi_get_devices_by_distance(selected_devices, max_devices_per_pe, tcurr);
            for (int i = 0; i < max_devices_per_pe; i++) {
                if (selected_devices[i] == -1) {
                    break;
                }
                found_devices++;
                INFO(NVSHMEM_INIT,
                     "NVSHMEM_ENABLE_NIC_PE_MAPPING = 0, device %d setting dev_id = %d", i,
                     selected_devices[i]);
            }
        }

        /* setting n_devices to 0 is the transports way of
         * letting us know it's managing devices internally.
         */
        if (tcurr->n_devices > 0 && selected_devices[0] == -1) {
            NVSHMEMI_ERROR_JMP(status, NVSHMEMX_ERROR_INTERNAL, out, "No devices selected.\n");
        }

        status = tcurr->host_ops.connect_endpoints(tcurr, selected_devices, found_devices);
        NVSHMEMI_NZ_ERROR_JMP(status, NVSHMEMX_ERROR_INTERNAL, out, "connect EPS failed \n");
        status = nvshmemi_boot_handle.barrier(&nvshmemi_boot_handle);
        NVSHMEMI_NZ_ERROR_JMP(status, NVSHMEMX_ERROR_INTERNAL, out, "barrier failed \n");

        status = nvshmemi_update_device_state();
    }

out:
    return status;
}
```

Now, it sees only one NIC, it makes tcurr->n_devices value always 1 ! Hence, in all cases, code flow will go in if code block. It will never check for else if or else block of the code; apparently, it can make NVSHMEM_ENABLE_NIC_PE_MAPPING environment variable irrelevant !

```cpp
        // assumes symmetry of transport list at all PEs
        if (tcurr->n_devices <= 1) {
            /* return the index of the first available device.
             * -1 if no devices found.
             */
            selected_devices[0] = tcurr->n_devices - 1;
            found_devices++;
        }
```

vimalKeshu:
@smarterclayton I think I am able to find the reference code from NVSHMEM source which clearly indicates that mype_node and npes_node values derived from nvshmemi_boot_handle.pg_rank and nvshmem_boot_handle.pg_size !

Look for below lines in code snippet:
```cpp
    mype = nvshmemi_boot_handle.pg_rank;
    npes = nvshmemi_boot_handle.pg_size;
    
    ....
    
    nvshmemi_boot_handle.mype_node = mype_node;
    nvshmemi_boot_handle.npes_node = npes_node;
    
```

Code snippet:
```cpp
int nvshmemi_bootstrap(int flags, nvshmemx_init_attr_t *nvshmem_attr) {
    int status = 0;
    uint64_t myHostHash = 0;
    uint64_t *hostHash = NULL;
    int mype, npes;
    int mype_node = 0, npes_node = 0;
    int num_nodes;

    bootstrap_attr_t attr = {};
    status = bootstrap_set_bootattr(flags, nvshmem_attr, (nvshmem_attr) ? &attr : NULL);
    NVSHMEMI_NZ_ERROR_JMP(status, NVSHMEMX_ERROR_INTERNAL, out, "bootstrap_set_bootattr failed \n");
    status = bootstrap_init(flags, &attr, &nvshmemi_boot_handle);
    NVSHMEMI_NZ_ERROR_JMP(status, NVSHMEMX_ERROR_INTERNAL, out, "bootstrap_init failed \n");
    if (nvshmemi_default_session == nullptr) {
        nvshmemi_default_session = (nvshmemi_session_t *)calloc(sizeof(nvshmemi_session_t), 1);
        NVSHMEMI_NULL_ERROR_JMP(nvshmemi_default_session, status, NVSHMEMX_ERROR_INTERNAL, out,
                                "Unable to allocate memory for nvshmem session");
        nvshmemi_default_session->bootstrap = &nvshmemi_boot_handle;
    }

    mype = nvshmemi_boot_handle.pg_rank;
    npes = nvshmemi_boot_handle.pg_size;
    myHostHash = getHostHash();
    hostHash = (uint64_t *)malloc(sizeof(uint64_t) * npes);
    status = nvshmemi_boot_handle.allgather((void *)&myHostHash, (void *)hostHash, sizeof(uint64_t),
                                            &nvshmemi_boot_handle);
    NVSHMEMI_NZ_ERROR_JMP(status, NVSHMEMX_ERROR_INTERNAL, out,
                          "allgather of host hashes failed \n");
    nvshmemi_host_hashes = (uint64_t *)malloc(sizeof(uint64_t) * npes);
    memcpy(nvshmemi_host_hashes, hostHash, sizeof(uint64_t) * npes);
    for (int i = 0; i < npes; i++) {
        if (nvshmemi_host_hashes[i] == myHostHash) {
            if (i == mype) mype_node = npes_node;
            npes_node++;
        }
    }
    nvshmemi_boot_handle.mype_node = mype_node;
    nvshmemi_boot_handle.npes_node = npes_node;

    // Check for same number of PEs on every node
    // Use myHostHash value to indicate a given PE has been counted already
    // This overwrites the hostHash values, but we don't them past this point

    for (int i = 0; i < npes; i++) {
        // If this host's node hasn't been counted yet, count it
        if (hostHash[i] != myHostHash) {
            const uint64_t peer_hash = hostHash[i];
            int npes_peer_node = 0;
            for (int j = i; j < npes; j++) {
                if (peer_hash == hostHash[j]) {
                    npes_peer_node++;
                    hostHash[j] = myHostHash;
                }
            }
            if (npes_peer_node != npes_node) {
                char hostname[1024];
                nvshmemu_gethostname(hostname, 1024);

                NVSHMEMI_ERROR_JMP(
                    status, NVSHMEMX_ERROR_INTERNAL, out,
                    "NVSHMEM requires the same number of PEs on all nodes (%d PEs on %s)\n",
                    npes_node, hostname);
            }
        }
    }

    nvshmem_nvtx_set_thread_name(mype);

    /* Set pe distribution type. First check for round robin distribution. Then check for block
     * distribution. */
    nvshmemi_pe_dist = NVSHMEMI_PE_DIST_MISC;

    if (npes_node != 0) {
        if (npes % npes_node != 0) goto out;
        num_nodes = npes / npes_node;
    } else {
        NVSHMEMI_ERROR_JMP(
            status, NVSHMEMX_ERROR_INTERNAL, out,
            "NVSHMEM hit the error of division by zero: npes_node == 0 what happen!?\n");
    }

    for (int i = 0; i < num_nodes; i++) {
        for (int j = 0; j < npes_node; j++) {
            if (nvshmemi_host_hashes[i * npes_node] != nvshmemi_host_hashes[i * num_nodes + j])
                goto check_roundrobin_dist;
        }
    }
    nvshmemi_pe_dist = NVSHMEMI_PE_DIST_BLOCK;
    INFO(NVSHMEM_INIT, "PE distribution has been identified as NVSHMEMI_PE_DIST_BLOCK");
    goto out;

check_roundrobin_dist:
    for (int i = 0; i < npes_node; i++) {
        for (int j = 0; j < num_nodes; j++) {
            if (nvshmemi_host_hashes[j * npes_node] != nvshmemi_host_hashes[i * num_nodes + j])
                goto out;
        }
    }
    nvshmemi_pe_dist = NVSHMEMI_PE_DIST_ROUNDROBIN;
    INFO(NVSHMEM_INIT, "PE distribution has been identified as NVSHMEMI_PE_DIST_ROUNDROBIN");

out:
    nvshmemi_device_state.pe_dist = nvshmemi_pe_dist;
    if (hostHash) free(hostHash);
    return status;
}
```

vimalKeshu:
The crux of the any solution to this problem is to pass NVSHMEM_HCA_LIST environment variable such a way that each device process only sees its nearest NIC.

Now,

1. this can be handled from deepEP code itself; as per existing PR.
2. or, it can be handled from parent process who uses deepEP lib in its code like vLLM. 

I see in both the cases, NVSHMEM_ENABLE_NIC_PE_MAPPING environment variable is not required.

deep_ep/buffer.py
```diff
@@ -135,6 +137,36 @@ def all_gather_object(obj):
         self.runtime.sync(device_ids, ipc_handles, root_unique_id)
         assert self.runtime.is_available()
 
+    def _setup_device_hca_mapping(self):
```
vimalKeshu, line 140: @smarterclayton  This issue can be handled in different ways: https://github.com/deepseek-ai/DeepEP/pull/466#issuecomment-4131128115

I think it would be good idea to implement `_setup_device_hca_mapping` method in parent process who underlying uses deepEP lib. 

`_setup_device_hca_mapping` method can be added in DeepEP inter node test, [test_internode.py](https://github.com/deepseek-ai/DeepEP/blob/main/tests/test_internode.py) and it can be called under [test_loop()](https://github.com/deepseek-ai/DeepEP/blob/main/tests/test_internode.py#L316)

That's way, it can provide strong demonstration to users of deepEP to add `_setup_device_hca_mapping` method into their parent process who uses DeepEP lib in case if they experience similar issue.

keonjang-google:
Thanks @vimalKeshu for the suggestions. We will try without setting NVSHMEM_ENANBLE_NIC_PE_MAPPING and get back to you on that. 

Having said that, I would still like to suggest to put this in the DeepEP rather than other parts. 
1. Unless one specifies the environmental variable, it has 0 side effects to DeepEP. 
2. It's not just vllm that needs to do this. SGLang, Nemo, also supports DeepEP and and depending on their environment they will need solution to manually map GPUs to NICs.

vimalKeshu:
hi @keonjang-google got you; make sense. thank you.

jiiwang:
> hi @keonjang-google got you; make sense. thank you.

Hi @vimalKeshu. We tested without setting NVSHMEM_ENANBLE_NIC_PE_MAPPING, and the results show that it is not required. We will update the PR soon. Thanks for your suggestion.

PR #470, opened 2025-10-29
Fix assert when probs are empty.

bnellnm:
I hit this assert when running `deepseek-ai/DeepSeek-Coder-V2-Lite-Instruct` and the tokens/probs assigned to one rank came up empty.

cc @Autumn1998 , @jershi425


jershi425:
@bnellnm Thanks for the fix!

PR #492, opened 2025-11-13
[Feature] Per Expert Overlap (PEO)

k-ling3:
## Background

The insight of this work is that we group the experts, allowing the communication of some experts to overlap with the computation of other experts. We call this approach Per Expert Overlap (PEO). Compared to existing methods, our approach has the following advantages:

### 1. Performance:
- **Compared to Non-overlap**, PEO performs better at all batch sizes.
  - For the DPSK model: 
    - At batch size 4, PEO achieves an 11% improvement.
    - At batch size 128, PEO achieves a 31% improvement. The larger the batch size, the more significant the gain.
  - For the QWEN model, PEO achieves up to a 51% improvement.

- **Compared to PR 390** (https://github.com/deepseek-ai/DeepEP/pull/390), PEO also performs better.

### 2. Usability
- **Compared to PR 390**, PEO only modifies DeepEP and does not change DeepGEEM, making it easier to use.

In short, during the dispatch phase, we change the order of communication (by modifying DeepEP) to allow some experts to receive tokens first. During the GEMM phase, we change the order of computation (by modifying how the inference engine calls DeepGEMM) to allow some experts to compute first. In the combine phase, we let some experts send tokens first. Overall, this allows the communication of some experts to overlap with the computation of others.

---

## Design

In the original DeepEP, each communication unit consists of `num_experts` or `num_local_experts` experts. That is, during the dispatch phase, each rank sends tokens to `num_experts` experts. During the combine phase, each rank sends tokens from `num_local_experts` experts to `num_ranks` ranks.

This solution modifies DeepEP by dividing the experts into `num_rounds` groups, and the communication is divided into `num_rounds` rounds.

- **During the dispatch phase**, in each round, each rank sends tokens to `num_experts // num_rounds` experts.
- **During the combine phase**, in each round, each rank sends tokens from `num_local_experts // num_rounds` local experts to `num_ranks` ranks.

The process is shown as follows:

Due to differences in model parameters, deployment scale, and batch size, this solution allows the following adjustable parameters to achieve the best overlap effect in different scenarios:

### Parameters for Overlap:
- **Overlap method**
  - We tested different overlap methods and found they have different effects. Consider the following options:
    - **overlap-1**: After all dispatch sends are completed, then perform dispatch recv and GEMM.
      <div style="text-align:center;">
        overlap-1 diagram
      </div>
    - **overlap-2**: Immediately after each dispatch send, perform recv + gemm.
      <div style="text-align:center;">
        overlap-2 diagram
      </div>
    - **overlap-3**: Immediately after each dispatch send, perform recv + gemm, and allow DeepEP's send and recv to overlap.
       <div style="text-align:center;">
        overlap-2 diagram
      </div>
    - **overlap-4**: No overlap between dispatch and GEMM.
      <div style="text-align:center;">
        overlap-2 diagram
      </div>

- **num_rounds**: Number of rounds for splitting dispatch/combine.
- **deepep_send_num_sms**: Number of SMs used for dispatch/combine send.
- **deepep_recv_num_sms**: Number of SMs used for dispatch/combine recv.
- **up_deepgemm_num_sms**: Number of SMs used for UP GEMM.
- **down_deepgemm_num_sms**: Number of SMs used for DOWN GEMM.

---

### Configuration:
- H20, EP16, QWEN, DPSK

### Comparison Methods:
- non-overlap
- PR 390 (https://github.com/deepseek-ai/DeepEP/pull/390) (SBO)

### Conclusion:
For both DPSK and QWEN, the overlap performance is the best at almost all batch sizes. For DPSK, PEO achieves a maximum of 31% improvement at batch size 128. For QWEN, PEO achieves a maximum of 50% improvement at batch size 16.

### DIFF
--- a/csrc/deep_ep.cpp
+++ b/csrc/deep_ep.cpp
@@ -1712,14 +1712,25 @@ std::tuple<torch::Tensor, std::optional<EventHandle>, std::optional<std::functio
     bool zero_copy,
     bool async,
     bool return_recv_hook,
-    bool use_expert_overlap, int num_rounds, int send_round_id, int send_num_sms, int recv_num_sms, bool hook_use_comm_stream,
+    bool use_expert_overlap, int num_rounds, int send_round_id, int send_num_sms, int recv_num_sms, bool hook_use_comm_stream, bool is_x_in_round,
     const std::optional<torch::Tensor>& out) {
 #ifndef DISABLE_NVSHMEM
     EP_HOST_ASSERT(low_latency_mode);
+    
+    int num_local_experts_per_round = -1;
+    if (is_x_in_round) {
+        EP_HOST_ASSERT(use_expert_overlap and "is_x_in_round is only supported when use_expert_overlap is enabled");
+    }
+    if (use_expert_overlap) {
+        EP_HOST_ASSERT(num_rounds >= 1 && num_experts / num_ranks % num_rounds == 0 && return_recv_hook);
+        if (send_num_sms == -1) send_num_sms = num_device_sms;
+        if (recv_num_sms == -1) recv_num_sms = num_device_sms;
+        num_local_experts_per_round = num_experts / num_ranks / num_rounds;
+    }
 
     // Tensor checks
     EP_HOST_ASSERT(x.dim() == 3 and x.is_contiguous() and x.scalar_type() == torch::kBFloat16);
-    EP_HOST_ASSERT(x.size(0) == num_experts / num_ranks);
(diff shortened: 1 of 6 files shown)


rubbberrabbit:
Hi, to achieve Parameters for Overlap, should we modify in sglang forward and launch multi time gemm( to caculate different expert group ) ?

k-ling3:
> Hi, to achieve Parameters for Overlap, should we modify in sglang forward and launch multi time gemm( to caculate different expert group ) ?

Yes. We need to modify the inference engine to launch multiple GEMM operations for different expert groups. You can refer to SGLang's PR for details: https://github.com/sgl-project/sglang/pull/13442

PR #504, opened 2025-11-27
Optimization of Hybrid-EP

Autumn1998:
1. Add topo-detection for RDMA 
2. Optimize the perf in small hidden/large EP
3. Use ~/.deepep/hybrid_ep/jit as the jit path
4. Group 2 sync into 1, and move sync point before the dispatch
5. dynamic seq len for RDMA
6. fix some bugs


dongwang4096:
Good job! It seems that there is an error in the position of the brackets, in topo_detection.cuh : Line 330-332 .

Autumn1998:
> Good job! It seems that there is an error in the position of the brackets, in topo_detection.cuh : Line 330-332 .

Fixed, this is also a bug on the old nccl version..

PR #514, opened 2025-12-07
Fix nvshmem_host_lib name in setup

Edenzzzz:
When installing nvshmem 3.3.20 via pip, the name of the file is `libnvshmem_host.so.3` instead of `libnvshmem_host.so`, so I guess it's more robust to use `get_nvshmem_host_lib_name`
image


polarstormx:
This has been addressed in #627 (commit 723716f), so closing as resolved.

Issue #516, opened 2025-12-10
Does deepep-tests scripts support nsys profile?


fishautumn:
it uses kineto to capture kernel durations. does it still work along with `nsys profile -t cuda`?

if it doesn't, can we add a command line option or some env variable to by-pass kineto data capturing and calculation, so `nsys profile` can work?

polarstormx:
Current main already provides `EP_USE_NVIDIA_TOOLS=1` to bypass the internal Kineto profiling, as documented in the README.

PR #534, opened 2025-12-25
Add DEEPEP_SKIP_KINETO env var to bypass kineto for nsys profiling

yurekami:
## Summary
- Add `DEEPEP_SKIP_KINETO` environment variable to bypass kineto profiling
- When set to `1`, uses simple CUDA event timing instead of torch.profiler
- Allows using external profilers like `nsys profile -t cuda` without conflicts

## Usage
```bash
DEEPEP_SKIP_KINETO=1 nsys profile -t cuda python tests/test_internode.py
```

## Background
The test scripts use kineto (torch.profiler) for capturing kernel durations. This can conflict with `nsys profile` when users want to analyze the kernels with NVIDIA Nsight Systems.

With this change, users can set `DEEPEP_SKIP_KINETO=1` to disable kineto and use nsys freely.

## Test plan
- [ ] Run tests with `DEEPEP_SKIP_KINETO=1` and verify they complete
- [ ] Run `nsys profile -t cuda` with this env var set

Addresses #516

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -177,6 +177,28 @@ def bench_kineto(fn,
                  trace_path: Optional[str] = None,
                  barrier_comm_profiling: bool = False,
                  num_kernels_per_period: int = 1):
+    # Skip kineto profiling when DEEPEP_SKIP_KINETO=1 is set
+    # This allows using external profilers like `nsys profile` without conflicts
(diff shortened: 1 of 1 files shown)


polarstormx:
Current main already provides `EP_USE_NVIDIA_TOOLS=1` to bypass the internal Kineto profiling, as documented in the README.

PR #538, opened 2025-12-25
Use atomicMax instead of atomicAdd for wait_recv_cost_stats

### LINKED ISSUE #473: `dispatch_wait_recv_cost_stats` is cumulating time from each warp, when instead computing a max across warps would be more helpful
```
atomicAdd(reinterpret_cast<unsigned long long*>(dispatch_wait_recv_cost_stats + src_rank), wait_recv_cost);
```
This op adds the wait time as seen by a given warp for a given src_rank. This time is being added by each warp. 

I wonder what is the utility of this metric? Each warp is waiting in parallel. Are we trying to infer slow ranks through this metric? 

More interesting would be the max across warps since that would indicate the actual wait time from a src_rank. 

Then we could also get a max of this max across src_ranks to actually infer the pure network comms latency of the recv kernel.

yurekami:
## Summary
- Changes `atomicAdd` to `atomicMax` for `dispatch_wait_recv_cost_stats` and `combine_wait_recv_cost_stats`
- When warps wait in parallel, summing wait times produces misleading metrics
- Max captures the actual bottleneck latency for performance diagnostics

## Changes
- `csrc/kernels/internode_ll.cu:418` - dispatch stats
- `csrc/kernels/internode_ll.cu:973` - combine stats

## Rationale
As noted in #473, when multiple warps receive data from the same source rank in parallel, using `atomicAdd` to sum their wait times doesn't provide meaningful insights. Using `atomicMax` instead captures the maximum wait time observed, which:
1. Identifies the actual communication bottleneck
2. Helps diagnose slow compute ranks
3. Provides actionable latency metrics for tools like DeepXTrace

Fixes #473

### DIFF
--- a/csrc/kernels/internode_ll.cu
+++ b/csrc/kernels/internode_ll.cu
@@ -415,7 +415,7 @@ LOW_LATENCY_DISPATCH_RECV:
             if (cumulative_local_expert_recv_stats != nullptr)
                 atomicAdd(cumulative_local_expert_recv_stats + local_expert_idx, num_recv_tokens);
             if (dispatch_wait_recv_cost_stats != nullptr)
-                atomicAdd(reinterpret_cast<unsigned long long*>(dispatch_wait_recv_cost_stats + src_rank), wait_recv_cost);
+                atomicMax(reinterpret_cast<unsigned long long*>(dispatch_wait_recv_cost_stats + src_rank), wait_recv_cost);
         }
         asm volatile("bar.sync %0, %1;" ::"r"(warp_group_id + 2), "r"(num_warps_per_group * 32));
         num_recv_tokens = shared_num_recv_tokens[warp_group_id];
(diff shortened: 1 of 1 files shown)


alpha-baby:
The use of the `atomicAdd` method is deliberately chosen because we need to measure numerous instances of dispatch and combine communication, accumulating all communication times to analyze—from a statistical perspective—whether any slow areas exist. We cannot analyze which rank might be slow based solely on a single communication instance. You could attempt to identify slow ranks through a single collective communication operation, but this would be inaccurate, as network jitter and GPU computing power fluctuations are consistently present.

yurekami:
Thanks for the explanation. The sum is the intended statistic: the per-rank aggregate over many dispatch/combine rounds is what the slow-rank analysis needs, and a per-call max would lose it. Closing.

PR #591, opened 2026-03-12
Hybrid ep nixl2

QiaoK:
# Pull Request: NIXL Integration for Hybrid-EP Multi-Node Communication

## Summary

This PR integrates **NIXL** (NVIDIA Inter-node eXchange Library) as an alternative inter-node communication backend for Hybrid-EP, alongside the existing DOCA/RDMA path. NIXL uses UCX for GPU-to-GPU communication and does not require the NCCL submodule or DOCA SDK, simplifying deployment in environments where DOCA is unavailable.

## Motivation

- **DOCA build complexity**: The DOCA path requires the NCCL submodule, DOCA SDK, and `doca_gpunetio_device.h`, which often fails with "No rule to make target" when DOCA is not installed.
- **NIXL availability**: NIXL provides a lighter-weight inter-node path using UCX and etcd, suitable for clusters without DOCA.
- **Build flexibility**: Users can choose NIXL or DOCA at build time via `USE_NIXL=1`.

### 1. NIXL Coordinator and Connector

- **`NIXLCoordinator`** (`internode_nixl.cuh` / `internode_nixl.cu`): Implements `allocate_buffers()`, `destroy()`, `grow_buffer_config()`, and `update_config()` using the NIXL connector.
- **`HybridEP_NIXLConnector`** (`nixl_connector.h` / `nixl_connector.cu`): Ported from hybrid-ep; manages NIXL agent lifecycle, buffer registration, peer exchange via etcd, UCX wireup, and GPU context building.
- **Backend integration** (`hybrid_ep_backend.cuh`, `nixl_types.h`, `nixl_warp.cuh`): Added `USE_NIXL`-guarded paths for `dispatch_gpu_nixl_ctx` and `combine_gpu_nixl_ctx` in dispatch/combine kernels.

### 2. Build System (`setup.py`)

- **Dual-path build**: When `HYBRID_EP_MULTINODE=1`:
  - **NIXL path** (`USE_NIXL=1`): Compiles `internode_nixl.cu` and `nixl_connector.cu`; links against NIXL, UCX, mlx5, ibverbs; skips NCCL/DOCA build.
  - **DOCA path** (`USE_NIXL=0`): Compiles `internode.cu`; builds NCCL submodule; uses DOCA objects for JIT.
- **`.use_nixl` fallback**: If `USE_NIXL` is not set but `.use_nixl` exists in the project root, the NIXL path is used (helps when pip build isolation does not propagate env vars).
- **NCCL build failure handling**: When the DOCA path fails (e.g. missing `doca_gpunetio_device.h`), a clear error suggests using `USE_NIXL=1` or creating `.use_nixl`.
- **Build logging**: Prints `use_nixl=...` and `-> NIXL path` or `-> DOCA path` during build.

### 3. JIT Compiler (`compiler.cu`)

- **USE_NIXL support**: When built with `USE_NIXL`, JIT-compiled inter-node kernels use NIXL libs instead of DOCA objects.
- **Runtime logging**: Prints `[HybridEP JIT] Inter-node: using NIXL` or `using DOCA/RDMA` when the JIT compiler initializes.

### 4. Runtime Logging (`hybrid_ep.cu`)

- **Init**: Prints `[HybridEP] Inter-node: using NIXL (node_rank=X local_rank=Y)` or `using DOCA/RDMA` when multinode coordinators are initialized.
- **allocate_buffers**: Prints `[HybridEP] NIXL allocate_buffers (num_of_nodes=X)` or `DOCA allocate_buffers` for debugging.

### 5. Documentation

- **README_Hybrid-EP.md**:
  - NIXL vs DOCA installation instructions.
  - Troubleshooting for `doca_gpunetio_device.h` error.
  - Two-node `hybrid_ep_test` configuration (`WORLD_SIZE`, `RANK`, `MASTER_ADDR`, etc.).
  - Single-machine simulation with `NUM_OF_HYBRID_EP_RANKS_PER_NVLINK_DOMAIN`.

### 6. Other Fixes

- **JIT race fix**: Process-specific JIT cache dir (`.../jit/proc-{pid}`) to avoid races when multiple processes compile concurrently.
- **Device API v2 / FP8**: Flag semantics, stride alignment, and GPU context struct updates for NIXL compatibility.

## Build Instructions

**NIXL path (recommended when DOCA is unavailable):**
```bash
export HYBRID_EP_MULTINODE=1
export USE_NIXL=1
export NIXL_HOME=/usr/local/nixl  # Adjust if needed
export TORCH_CUDA_ARCH_LIST="9.0 10.0"
pip install .
```

**DOCA path (when DOCA SDK is available):**
```bash
export HYBRID_EP_MULTINODE=1
export RDMA_CORE_HOME=/path/to/rdma-core
pip install .
```

**Fallback when env does not propagate:**
```bash
touch .use_nixl
HYBRID_EP_MULTINODE=1 pip install .
```

## Testing

- **Single-node**: `python tests/test_hybrid_ep.py --num-processes 4` (unchanged; NIXL not used).
- **Multi-node**: Set `WORLD_SIZE=2`, `RANK=0` or `1`, `MASTER_ADDR`, `MASTER_PORT` on each node; run `python tests/test_hybrid_ep.py --num-processes 4` on both.
- **Single-machine simulation**: `NUM_OF_HYBRID_EP_RANKS_PER_NVLINK_DOMAIN=2 python tests/test_hybrid_ep.py --num-processes 4` to simulate 2 nodes on 4 GPUs.

## Dependencies

- **NIXL path**: NIXL library, UCX, mlx5, ibverbs. Optional: etcd (NIXL_ETCD_ENDPOINTS) for multi-node.
- **DOCA path**: NCCL submodule, DOCA SDK, rdma-core (unchanged).

## Backward Compatibility

- `USE_NIXL=0` (default) preserves the DOCA path; no behavior change for existing DOCA users.
- NIXL code is fully guarded by `#ifdef USE_NIXL`; no NIXL symbols are linked when using DOCA.


csrc/hybrid_ep/backend/nixl_warp.cuh
```diff
@@ -0,0 +1,223 @@
...
+         typename SMEM_TYPE,
+         int NUM_OF_STAGES,
+         int HIDDEN_DIM,
+         int NUM_OF_TOKENS_PER_CHUNK,
+         int NUM_OF_EXPERTS_PER_RANK,
+         int NUM_OF_RANKS_PER_NODE,
+         int NUM_OF_NODES,
+         int NUM_OF_BLOCKS,
+         bool FORWARD_DISPATCH>
+inline __device__ void N2N_warp_group_device_function(const int node_rank,
```
Autumn1998: The content in this file seems to already exist in the backend — it's best to keep it in one place only. I think nixl_wrapper.cuh/nixl_types.h, should be moved back into the backend. If we want to refactor the hybrid_ep_backend.cuh to make it cleaner, that can be done in a follow-up PR.

csrc/hybrid_ep/backend/hybrid_ep_backend.cuh
```diff
@@ -435,31 +556,50 @@ struct dispatch_kernel_param_t{
...
   uint64_t* expected_rdma_flag_value;
-  uint32_t* expected_intra_node_flag_value;
+  uint32_t* expected_intra_node_flag_value; // Need 2 expected values for different parity for dispatch kernel.
+  uint32_t* expected_permute_flag_value;
+  uint32_t* intra_node_flag_parity;
   int local_rank;
   int node_rank;
   // The number of token output by attn layer on a rank/GPU.
   int num_of_tokens_per_rank;
+#ifdef HYBRID_EP_BUILD_MULTINODE_ENABLE
```
Autumn1998: It's better to avoid using macros in param_t. I think this could be changed to 3 co-existing void parameters, and then at the actual call reinterpret_cast where arguments are passed in.

The reason is as follows: param_t is the only struct that we initialize on the runtime side but need to pass into the JIT kernel. Consider the case where we build with HYBRIDEP_ENABLE_MULTINODE enabled at compile time, but then use that jit build for single-node communication. In this case, the JIT compilation will not include the multi-node related dependencies.

So at runtime, the struct contains fields gated by the RDMA-related macros, but since those macros are not defined during JIT compilation, the two sides end up with different definitions of param_t — we would be passing a struct built with HYBRIDEP_ENABLE_MULTINODE into a kernel that was compiled without it. This will often still work, but it is fundamentally undefined behavior and leaves a hidden risk.

csrc/hybrid_ep/buffer/internode_nixl.cu
```diff
@@ -0,0 +1,199 @@
...
+        buffer_config.num_of_nodes,
+        buffer_config.num_of_ranks_per_node,
+        buffer_config.num_of_blocks_dispatch_api,
+        buffer_config.num_of_blocks_combine_api,
+        dispatch_buffers,
+        combine_buffers);
+
+    auto torch_distributed = pybind11::module_::import("torch.distributed");
+    torch_distributed.attr("barrier")(this->process_group);
+    std::this_thread::sleep_for(std::chrono::seconds(3));
```
Autumn1998: why we need this sleep here ?

csrc/hybrid_ep/buffer/nixl_connector.cu
```diff
@@ -0,0 +1,710 @@
...
+    NIXL_LOG("  [Rank %d] connectRanks: Building GPU contexts\n", rank_uuid);
+    _nixl_build_gpu_contexts(num_channels, num_channels);
+
+    connected_ranks = ranks_to_connect;
+    connected = true;
+
+    NIXL_LOG("  [Rank %d] connectRanks: Successfully connected to %zu ranks\n", rank_uuid, connected_ranks.size());
+}
+
+void HybridEP_NIXLConnector::disconnectRanks(const std::vector<int>& remote_rank_uuids) {
```
Autumn1998, line 345: When destructing the nixl-internode buffer, we trigger the reset of the connector, but this `disconnectRanks` doesn't seem to be used? If the primary purpose of the nix connector is to perform a connect operation on each memory buffer (perhaps we could add a brief comment in the header to describe what this connect is actually doing), then should we call disconnect first before the destructor completes?

csrc/hybrid_ep/jit/compiler.cu
```diff
@@ -49,6 +59,24 @@ NVCCCompiler::NVCCCompiler(std::string base_path, std::string comm_id):
...
+    fprintf(stderr, "[HybridEP JIT] Inter-node: using NIXL\n");
+    flags += " -DUSE_NIXL";
+    std::string nixl_home = get_env("NIXL_HOME");
+    if (nixl_home.empty()) nixl_home = "/usr/local/nixl";
+    std::string ucx_home = get_env("UCX_HOME");
+    if (ucx_home.empty()) ucx_home = "/usr";
+    include += " -I" + nixl_home + "/include ";
+    include += " -I" + nixl_home + "/include/gpu/ucx ";
+    include += " -I" + ucx_home + "/include ";
+    std::string doca_home = get_env("DOCA_HOME");
```
Autumn1998: What does DOCA_HOME refer to here? I couldn't find any related instructions in the README. Also, what is the relationship between DOCA_HOME and DOCA_PATH? Do they share the same dependencies?

docs/README_Hybrid-EP.md
```diff
@@ -151,11 +151,24 @@ export TORCH_CUDA_ARCH_LIST="9.0 10.0"  # Adjust based on your GPU architecture
...
-For multi-node support with RDMA, additional configuration is required, make sure RDMA core libraries are properly installed and the path points to the directory containing the RDMA headers and libraries.
+#### Multi-node NIXL Installation (recommended when NIXL is available)
+For multi-node support with NIXL (no DOCA/NCCL required):
 
 ```bash
 export HYBRID_EP_MULTINODE=1
+export USE_NIXL=1
+export NIXL_HOME=/usr/local/nixl  # Adjust if NIXL is installed elsewhere
+export UCX_HOME=/usr              # Adjust if UCX is installed elsewhere
+export TORCH_CUDA_ARCH_LIST="9.0 10.0"  # Adjust based on your GPU architecture
```
Autumn1998, line 177: I also noticed the DOCA_HOME dependency in the code — is this not needed?

docs/README_Hybrid-EP.md
```diff
@@ -151,11 +151,24 @@ export TORCH_CUDA_ARCH_LIST="9.0 10.0"  # Adjust based on your GPU architecture
 pip install .
 ```
 
-#### Multi-node RDMA Installation
-For multi-node support with RDMA, additional configuration is required, make sure RDMA core libraries are properly installed and the path points to the directory containing the RDMA headers and libraries.
+#### Multi-node NIXL Installation (recommended when NIXL is available)
+For multi-node support with NIXL (no DOCA/NCCL required):
```
Autumn1998: Since HybridEP relies on the torch.distributed group, it is actually possible that NCCL may be used for some metadata communication. So this description is not entirely accurate.

docs/README_Hybrid-EP.md
Autumn1998: Additionally, I think we should explain the source of NIXL_HOME and UCX_HOME here in more detail — if we need to install NIXL, what is the recommended version? What is the link? Or even better, is there an installation guide for NIXL/UCX?

csrc/hybrid_ep/jit/compiler.cu
```diff
@@ -49,6 +59,24 @@ NVCCCompiler::NVCCCompiler(std::string base_path, std::string comm_id):
 #ifdef HYBRID_EP_BUILD_MULTINODE_ENABLE
     // Add the dependency of the inter-node jit
     flags += " -DHYBRID_EP_BUILD_MULTINODE_ENABLE";
+#ifdef USE_NIXL
+    fprintf(stderr, "[HybridEP JIT] Inter-node: using NIXL\n");
```
Autumn1998: I do not recommend printing logs during JIT compilation — is this debug info? During large-scale training, this part of the log could generate thousands of lines.

csrc/hybrid_ep/buffer/internode_nixl.cu
QiaoK: It was part of the debugging process. removed.

docs/README_Hybrid-EP.md
QiaoK: Revised comments.

csrc/hybrid_ep/jit/compiler.cu
QiaoK: Yes, it was for testing if NIXL was used during debugging. Now removed.

csrc/hybrid_ep/hybrid_ep.cuh
```diff
@@ -69,7 +70,11 @@ public:
 private:
   ExtendedMemoryAllocator remote_allocator;
 #ifdef HYBRID_EP_BUILD_MULTINODE_ENABLE
+#ifdef USE_NIXL
```
Autumn1998, line 73: Here we can unify the nixl coordinator and rdma coordinator under the same name (e.g., both called `rdma_coordinator,` or go a step further and rename the DOCA backend's rdma coordinator to DocaCoordinator), so that when we later call methods like `init`, `destroy`, `update`, etc., we no longer need to use USE_NIXL to differentiate between them.

tests/test_hybrid_ep.py
```diff
@@ -444,42 +444,47 @@ def test_hybrid_ep_benchmark(buffer: deep_ep.HybridEPBuffer, group: dist.Process
...
+                ref = TorchRef(
+                    ep_group=group,
+                    num_of_experts=NUM_OF_EXPERTS,
+                    num_of_ranks_per_node=NUM_OF_RANKS_PER_NODE,
+                )
+
+                test_hybrid_ep_correctness(buffer, ref, use_fp8)
+                test_hybrid_ep_benchmark(buffer, group, use_fp8, args.nsys_profile)
+        dist.barrier()
+    finally:
```
Autumn1998: try-catch can sometimes hide errors, so I wouldn't recommend modifying the code in the tests.

csrc/hybrid_ep/backend/hybrid_ep_backend.cuh
```diff
@@ -728,71 +717,64 @@ inline __device__ void N2N_warp_group_device_function(const int node_rank,
...
+          // (token, prob if forward_dispatch, scaling_factor if use_fp8).
+          unsigned long local_sub = 0;
+          unsigned long remote_sub = 0;
 
-          unsigned long buffer_sub_idx = 0;
-          
           {
-            size_t local_offset = token_idx * static_cast<int64_t>(HIDDEN_DIM) * sizeof(TOKEN_DATA_TYPE);
-            size_t remote_offset = token_idx * static_cast<int64_t>(HIDDEN_DIM) * sizeof(TOKEN_DATA_TYPE);
+            size_t local_offset = token_idx * HIDDEN_DIM * sizeof(TOKEN_DATA_TYPE);
```
Autumn1998: On the calculation of offset related to HIDDEN_DIM, we should use  static_cast<int64_t>() to avoid the overflow

csrc/hybrid_ep/backend/hybrid_ep_backend.cuh
```diff
@@ -5745,7 +5721,7 @@ public:
 
   /*private:
   // Rank within the current node/host.
-  int local_rank_;
```
Autumn1998, line 5748: There are a large number of similar, meaningless changes in this PR — just adding or removing spaces, or changing comments — that have no actual impact on the code but add noise to the review. I'd suggest running an AI scan to clean up these irrelevant changes. Of course, the root cause is that there's currently no formatter script in the branch; we'll add one in the future.

QiaoK: Reformatted to remove these errors.

csrc/hybrid_ep/backend/hybrid_ep_backend.cuh
QiaoK: Added the type casting.

tests/test_hybrid_ep.py
QiaoK: Removed try-catch.

csrc/hybrid_ep/hybrid_ep.cuh
QiaoK: Unified both names to be internode_coordinator.

csrc/hybrid_ep/backend/hybrid_ep_backend.cuh
```diff
@@ -8,50 +8,49 @@
 #include <cuda_bf16.h>
 #include <cuda/ptx>
 #ifdef HYBRID_EP_BUILD_MULTINODE_ENABLE
-#ifndef USE_NIXL
-#include "doca_gpunetio_host.h"
-#include "doca_gpunetio_device.h"
-#include "infiniband/verbs.h"
-#include "infiniband/mlx5dv.h"
-#else
+#ifdef USE_NIXL
```
Autumn1998: The changes to these headers are actually meaningless changes, they merely involve moving and repositioning some code.

csrc/hybrid_ep/backend/hybrid_ep_backend.cuh
```diff
@@ -249,13 +248,13 @@ template<int NUM_OF_STAGES,
          int NUM_OF_TOKENS_PER_CHUNK,
          int NUM_OF_EXPERTS_PER_RANK,
          int NUM_OF_RANKS_PER_NODE> 
-struct dispatch_kernel_dynamic_shared_memory_buffer_t<uint16_t, NUM_OF_STAGES, HIDDEN_DIM, NUM_OF_TOKENS_PER_CHUNK, NUM_OF_EXPERTS_PER_RANK, NUM_OF_RANKS_PER_NODE, 1, true>{
+struct dispatch_kernel_dynamic_shared_memory_buffer_t<uint8_t, NUM_OF_STAGES, HIDDEN_DIM, NUM_OF_TOKENS_PER_CHUNK, NUM_OF_EXPERTS_PER_RANK, NUM_OF_RANKS_PER_NODE, 1, false>{
```
Autumn1998: The change here also simply swaps the positions of `dispatch_kernel_dynamic_shared_memory_buffer_t<uint8_t ...> `and `dispatch_kernel_dynamic_shared_memory_buffer_t<uint16_t ...>.`

docs/README_Hybrid-EP.md
```diff
@@ -151,11 +151,28 @@ export TORCH_CUDA_ARCH_LIST="9.0 10.0"  # Adjust based on your GPU architecture
 pip install .
 ```
 
-#### Multi-node RDMA Installation
-For multi-node support with RDMA, additional configuration is required, make sure RDMA core libraries are properly installed and the path points to the directory containing the RDMA headers and libraries.
+#### Multi-node NIXL Installation (recommended when NIXL is available)
+For multi-node support with NIXL. NIXL replaces DOCA for inter-node GPU data transfers, so the DOCA SDK and NCCL submodule are not needed at build time. Note that NCCL may still be used at runtime by `torch.distributed` for collective metadata operations.
+
+**Prerequisites:**
+- **NIXL** ([ai-computing/nixl](https://github.com/ai-computing/nixl)) — GPU-aware inter-node communication library. Install from source; see the NIXL README for build instructions.
```
Autumn1998: This link will lead to 404 Error

docs/README_Hybrid-EP.md
```diff
@@ -151,11 +151,28 @@ export TORCH_CUDA_ARCH_LIST="9.0 10.0"  # Adjust based on your GPU architecture
...
+- **UCX** ([openucx/ucx](https://github.com/openucx/ucx)) — typically already present in NVIDIA container images or available via `apt install libucx-dev`. UCX v1.17+ is recommended.
```
Autumn1998: I think we should provide concrete examples of which images contain UCX.
Furthermore, in my testing, the environment for NIXL and UCX is very complex. If possible, it would be best to have a Dockerfile based on the PyTorch image.

setup.py
```diff
@@ -96,51 +101,87 @@ def get_extension_hybrid_ep_cpp():
...
+            compile_args["nvcc"].append("-DUSE_NIXL")
+            sources.extend([
+                "csrc/hybrid_ep/buffer/internode_nixl.cu",
+                "csrc/hybrid_ep/buffer/nixl_connector.cu",
+            ])
+            nixl_home = os.getenv("NIXL_HOME", "/usr/local/nixl")
+            ucx_home = os.getenv("UCX_HOME", "/usr")
+            nixl_include = os.path.join(nixl_home, "include")
+            nixl_gpu_include = os.path.join(nixl_home, "include/gpu/ucx")
+            nixl_lib = os.path.join(nixl_home, "lib/x86_64-linux-gnu")
```
Autumn1998: We hard-core x86 arch here, shall we add the support of the arm64  case?

csrc/hybrid_ep/backend/hybrid_ep_backend.cuh
QiaoK: Reverted reordering.

csrc/hybrid_ep/backend/hybrid_ep_backend.cuh
QiaoK: Restored template orders

docs/README_Hybrid-EP.md
QiaoK: Link fixed.

setup.py
QiaoK: Added arm64.

csrc/hybrid_ep/backend/hybrid_ep_backend.cuh
```diff
@@ -15,43 +15,42 @@
...
-#ifdef USE_NIXL
-// GPU-side context for NIXL RDMA transfers.
-// Each memory view is indexed by remote_idx * number of buffers per remote node + buffer index.
-// num_channels maps to UCX_RC_GDA_NUM_CHANNELS for QP channel distribution.
-
 struct dispatch_gpu_nixl_ctx {
-  nixlMemViewH local_mvh;                 // Local source buffers (token, prob, scaling_factor)
-  nixlMemViewH remote_data_mvh;           // Remote data buffers indexed by remote_idx
-  nixlMemViewH remote_signal_mvh;         // Remote signal buffers indexed by remote_idx
-  uint64_t *local_flag_counters;          // [num_remote_nodes] Local completion counters
```
Autumn1998, line 38: Does this changes only reorder the code without any substantive modifications?

QiaoK: Restored comments for these variables.

PR #594, opened 2026-03-23
ll dispatch: bulk put per expert to reduce RDMA overhead

Xuan-1998:
Stage dispatch tokens into per-expert contiguous regions in the send buffer, then issue one bulk RDMA put per remote expert instead of one per token. This reduces the number of RDMA operations and proxy channel round-trips for the low-latency dispatch path.

### DIFF
--- a/csrc/kernels/internode_ll.cu
+++ b/csrc/kernels/internode_ll.cu
@@ -251,7 +251,7 @@ __global__ __launch_bounds__(1024, 1) void dispatch(void* packed_recv_x,
             }
             asm volatile("bar.sync 1, %0;" ::"r"(num_threads));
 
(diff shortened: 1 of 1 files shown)


alpha-baby:
How much performance improvement does this provide? Are there specific test data?

thepointer1982 (commented):
Zu

Xuan-1998:
> How much performance improvement does this provide? Are there specific test data?

This pr got accidentally closed, but I saw ~50% performance gain on EFA

alpha-baby:
> > How much performance improvement does this provide? Are there specific test data?
> 
> This pr got accidentally closed, but I saw ~50% performance gain on EFA

Do you have specific test data?

Issue #629, opened 2026-05-10
EP_DISABLE_GIN=1 does not bypass NCCL Gin preflight check


DanielDanyang:
## Summary

DeepEP V2/elastic still fails during `ElasticBuffer` initialization with `EP_DISABLE_GIN=1` when NCCL reports no Gin capability for the communicator.

This blocks pure single-node intra-node EP runs.

## Reproduction

Environment used to reproduce:

```text
GPU: 8 x NVIDIA B300 SXM6 AC
CUDA compiler: /usr/local/cuda-13.2/bin/nvcc
PyTorch: 2.13.0.dev20260502+cu130
NCCL: nvidia-nccl-cu13 2.30.4
NVSHMEM: nvidia-nvshmem-cu13 3.4.5
```

Run a pure intra-node elastic EP smoke test:

```bash
export CUDA_VISIBLE_DEVICES=0,1
export MASTER_ADDR=127.0.0.1
export MASTER_PORT=29591
export NCCL_DEBUG=WARN
export EP_DISABLE_GIN=1
export EP_JIT_CACHE_DIR=/tmp/deepep_b300_jit_cache

python tests/elastic/test_ep.py \
  --num-processes 2 \
  --num-tokens 128 \
  --hidden 1024 \
  --num-topk 2 \
  --num-experts 16 \
  --allow-hybrid-mode 0 \
  --test-first-only \
  --skip-perf-test
```

## Actual Result

The original code fails in host-side initialization:

```text
csrc/kernels/backend/nccl.cu:87
(allow_hybrid_mode ? props.railedGinType : props.ginType) != NCCL_GIN_TYPE_NONE
```

On the reproduced system:

```text
props.ginType == NCCL_GIN_TYPE_NONE
props.railedGinType == NCCL_GIN_TYPE_NONE
```

PR #630, opened 2026-05-10
Fix V2 initialization for pure single-node intra EP

DanielDanyang:
## Summary

This PR fixes V2/elastic initialization for pure single-node intra-node EP runs where NCCL reports no Gin capability. (#629)

This keeps the existing `EP_DISABLE_GIN` policy: when the variable is set, DeepEP skips Gin context requests instead of forcing Gin availability during preflight initialization.

## Validation

Validated on 8 x NVIDIA B300 GPUs, CUDA 13.2, PyTorch `2.13.0.dev20260502+cu130`, and NCCL `2.30.4`.

```bash
export EP_DISABLE_GIN=1

python tests/elastic/test_ep.py \
  --num-processes 8 \
  --num-tokens 8192 \
  --hidden 7168 \
  --num-topk 8 \
  --num-experts 256 \
  --allow-hybrid-mode 0 \
  --test-first-only
```

Average across 8 ranks:

| Metric | Result | README SM100 EP8 reference |
| --- | ---: | ---: |
| Dispatch | 731.9 GB/s SU | 726 GB/s NVLink |
| Combine | 741.0 GB/s SU | 740 GB/s NVLink |

### DIFF
--- a/csrc/kernels/backend/nccl.cu
+++ b/csrc/kernels/backend/nccl.cu
@@ -81,17 +81,20 @@ NCCLSymmetricMemoryContext::NCCLSymmetricMemoryContext(const int64_t& nccl_comm,
     if (get_env<int>("EP_BUFFER_DEBUG"))
         printf("EP NCCL device communicator has %d allocated QPs\n", num_allocated_qps);
 
-    // Query NCCL supported Gin Type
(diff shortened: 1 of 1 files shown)


sphish:
Thanks！

Issue #621, opened 2026-05-06
DeepEP LL Combine Race on NVL72


elvircrn:
# Reproduction Steps

We setup the following test:

* single NVL72 node with 4 GB200s
* `topk` = 2
* only rank 0 has tokens going into dispatch, the rest are marked `-1`
* each tokens gets routed to expert 0 and 1
* both experts weights are 1
* batch size 256 or 1024

with the following code:

```
import os
os.environ['NVSHMEM_QP_DEPTH'] = '4096'

import torch, torch.distributed as dist, deep_ep

dist.init_process_group(backend='nccl')
rank, W = dist.get_rank(), dist.get_world_size()
torch.cuda.set_device(rank)

H, E = 7168, 256
N = 1024

for T in [256, 1024]:
    buf = deep_ep.Buffer(
        dist.new_group(list(range(W))),
        num_nvl_bytes=0,
        num_rdma_bytes=deep_ep.Buffer.get_low_latency_rdma_size_hint(T, H, W, E),
        low_latency_mode=True, num_qps_per_rank=E//W,
        allow_mnnvl=True, explicitly_destroy=True)

    torch.manual_seed(42)
    x = torch.randint(0, 2, (T, H), dtype=torch.bfloat16, device='cuda') if rank == 0 else         torch.zeros(T, H, dtype=torch.bfloat16, device='cuda')

    if rank == 0:
        ids = torch.zeros(T, 2, dtype=torch.int64, device='cuda')
        ids[:, 0] = 0; ids[:, 1] = 1
        w = torch.ones(T, 2, dtype=torch.float32, device='cuda')
        expected = (2.0 * x.float()).bfloat16()
    else:
        ids = torch.full((T, 2), -1, dtype=torch.int64, device='cuda')
        w = torch.zeros(T, 2, dtype=torch.float32, device='cuda')
        expected = torch.zeros(T, H, dtype=torch.bfloat16, device='cuda')

    n_wrong = 0
    for it in range(N):
        buf.clean_low_latency_buffer(T, H, E)
        torch.cuda.synchronize()
        dist.barrier()

        ex, _, h, _, hk = buf.low_latency_dispatch(
            x, ids, T, E, use_fp8=False, async_finish=False, return_recv_hook=True)
        hk()
        torch.cuda.synchronize()

        out = torch.zeros(T, H, dtype=torch.bfloat16, device='cuda')
        _, _, hk_c = buf.low_latency_combine(
[...]

PR #632, opened 2026-05-11
Fix: Add fence.proxy.async before TMA load in combine recv phase

elvircrn:
# Background

Correctness issues with repeated runs of combine at bath sizes higher than 256 were found and [issue](https://github.com/deepseek-ai/DeepEP/issues/621) was reported.

An investigation was done revealing doing __ldg on the `buffer` ptr instead of TMA make combine produce consistent results at bach size > 256 on NVL72. A plausible explanation for this is that switching to generic proxy resolves the issue.

Investigating further revealed that the local rank polls the generic proxy but then uses TMA to load data. Without `fence.proxy.async`, TMA can see stale data.

From [https://docs.nvidia.com/cuda/parallel-thread-execution/#async-proxy](url):

```
Accessing the same memory location across multiple proxies needs a cross-proxy fence. 
For the async proxy, fence.proxy.async should be used to synchronize memory between 
generic proxy and the async proxy.
```

With the fence in place, the test scripts made as part of [this](https://github.com/deepseek-ai/DeepEP/issues/621) report correct behavior.

# Fence Performance Impact

```
low-latency benchmark (median of 3 rounds, 1000 iters each)
  T=1024  H=7168  E=256  topk=8  ranks=4
  worst-case rank (max across ranks per round)

  strategies:
    random      = topk over random scores, different experts per token (cross-rank RDMA)
    random-same = topk over random scores, same K experts for all tokens (cross-rank RDMA)
    local-rand  = random offsets mapped to local rank experts (no RDMA, varied experts)
    local-same  = every token picks the same K local experts (no RDMA, hotspot)
    remote-rand = random experts, each from a different remote rank (all cross-rank RDMA)

strategy     fmt       cleg_on  cleg_off   delta     pct
                        (us)      (us)     (us)
------------ ------   -------  --------  ------  ------
random       bf16      178.1     170.1    +8.0   +4.7%
random       fp8       178.2     170.3    +7.9   +4.6%
random-same  bf16      646.8     639.0    +7.7   +1.2%
random-same  fp8       647.8     640.2    +7.6   +1.2%
local-rand   bf16      177.3     172.5    +4.8   +2.8%
local-rand   fp8       181.4     177.2    +4.2   +2.4%
local-same   bf16      656.2     653.2    +3.0   +0.5%
local-same   fp8       657.8     655.1    +2.7   +0.4%
remote-rand  bf16      223.7     215.8    +8.0   +3.7%
remote-rand  fp8       226.0     217.8    +8.2   +3.8%
```

Benchmark [script](https://gist.github.com/elvircrn/2e564cc0a53c3a72ab887480686a3456).

### DIFF
--- a/csrc/kernels/legacy/internode_ll.cu
+++ b/csrc/kernels/legacy/internode_ll.cu
@@ -1060,6 +1060,7 @@ LOW_LATENCY_COMBINE_RECV:
                             num_casted = (info >> 1) + (info & 1);
                         }
                         int num_tma_bytes = num_casted * kNumLogFMTPerWarpBytes + (num_decode_warps - num_casted) * kNumBF16PerWarpBytes;
+                        asm volatile("fence.proxy.async;");
                         tma_load_1d(
                             tma_ld_buffers[stage_idx], buffer + (kUseLogFMT ? kNumMetaBytes : 0), full_barriers[stage_idx], num_tma_bytes);
                         mbarrier_arrive_and_expect_tx(full_barriers[stage_idx], num_tma_bytes);


RayWang96:
Hello, I was also able to reproduce this issue on NVL72. The reason is that the TMA load gets issued before LDS has finished, which ends up corrupting shared memory. So we only `need fence.proxy.async.shared::cta;` to fix the issue, with almost no performance impact. Here is my PR https://github.com/deepseek-ai/DeepEP/pull/642.

PR #642, opened 2026-05-24
Add fence.proxy.async.shared::cta between mbarrier wait and TMA load.

RayWang96:
This fixes the issue mentioned in https://github.com/deepseek-ai/DeepEP/issues/621. Compared with the fix in https://github.com/deepseek-ai/DeepEP/pull/632, this version uses a CTA-scope fence, which resolves the problem while preserving performance.

The data race here comes from LDS and TMA loads, so we only need fence.proxy.async.shared::cta; to prevent the TMA load from being issued before LDS has completed.

Thanks to [elvircrn](https://github.com/elvircrn)'s script, I've tried my fix on NVL72, the result is correct and the performance is as good as the original implementation.

### DIFF
--- a/csrc/kernels/legacy/internode_ll.cu
+++ b/csrc/kernels/legacy/internode_ll.cu
@@ -1060,7 +1060,6 @@ LOW_LATENCY_COMBINE_RECV:
                             num_casted = (info >> 1) + (info & 1);
                         }
                         int num_tma_bytes = num_casted * kNumLogFMTPerWarpBytes + (num_decode_warps - num_casted) * kNumBF16PerWarpBytes;
-                        fence_view_async_shared();
                         tma_load_1d(
                             tma_ld_buffers[stage_idx], buffer + (kUseLogFMT ? kNumMetaBytes : 0), full_barriers[stage_idx], num_tma_bytes);
                         mbarrier_arrive_and_expect_tx(full_barriers[stage_idx], num_tma_bytes);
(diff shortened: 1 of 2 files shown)


elvircrn:
Might be cleaner to use:

```
__device__ __forceinline__ void tma_store_fence() {
    asm volatile("fence.proxy.async.shared::cta;");
}
```

from `utils.cuh`.

elvircrn:
Curious, does the issue revealed by my script go away with `fence.proxy.async.shared::cta`?

sphish:
@elvircrn Thank you for reporting this bug. 
> Investigating further revealed that the local rank polls the generic proxy but then uses TMA to load data. Without fence.proxy.async, TMA can see stale data.

For this issue you mentioned, the current approach does not strictly conform to the CUDA memory model, but based on some experiments we have done and our assumptions about the TMA implementation, we believe the current approach is safe on the existing architecture.

RayWang96:
> Might be cleaner to use:
> 
> ```
> __device__ __forceinline__ void tma_store_fence() {
>     asm volatile("fence.proxy.async.shared::cta;");
> }
> ```
> 
> from `utils.cuh`.

Here, using fence is unrelated to the TMA store, so for clarity, I will use the same naming as in [CUTLASS](https://github.com/NVIDIA/cutlass/blob/main/include/cutlass/arch/barrier.h#L728): void fence_view_async_shared()

RayWang96:
> Curious, does the issue revealed by my script go away with `fence.proxy.async.shared::cta`?

Yes, I tried your script, and after the fix there are no issues:

T=256: correct=1024/1024
T=1024: correct=1024/1024

PR #660, opened 2026-06-04
Fix PP TMA stage reuse race

### LINKED ISSUE #657: [DeepEP V2] tma_copy data race
https://github.com/deepseek-ai/DeepEP/blob/d4f41e4e93602a15e95f55f6ee8df8f1aaa0e4bb/deep_ep/include/deep_ep/impls/pp_send_recv.cuh#L88-L106

In the `tma_copy` function, the prefetch stage uses `tma_store_wait<kNumStages - 1>()` before reissuing a TMA load into the same shared memory buffer. With `kNumStages = 2`, this translates to `cp.async.bulk.wait_group 1`, which only waits until the number of pending store groups drops to ≤ 1 — meaning the current stage's store may still be in-flight when the next load overwrites the same smem buffer.

The pipeline executes as follows (with kNumStages = 2):

`iter 0, stage 0`: Load from global → stage0_smem. Wait mbarrier. Issue TMA store from stage0_smem → global; commit → group[A]. `Prefetch`: call tma_store_wait<1>().

At this point, group[A] (which reads from stage0_smem) is the only pending group, so wait_group 1 returns immediately without waiting for group[A] to complete.
A new tma_load_1d is then immediately issued into stage0_smem for the next iteration(iter=0+2=2).
Race: tma_store (group[A]) is still reading from stage0_smem while tma_load is writing new data into stage0_smem — this is a read-after-write hazard on the shared memory buffer, resulting in potential data corruption.

qescccczmr:
Fix a PP send/recv TMA copy race when reusing a shared-memory pipeline stage.
  This changes the reuse wait to `ptx::tma_store_wait()` / `wait_group 0`, ensuring all outstanding TMA
  stores complete before the same stage buffer is reused.
  Fixes #657.

### DIFF
--- a/deep_ep/include/deep_ep/impls/pp_send_recv.cuh
+++ b/deep_ep/include/deep_ep/impls/pp_send_recv.cuh
@@ -96,7 +96,7 @@ __device__ __forceinline__ void tma_copy(
         // Prefetch: wait until this stage's buffer is safe to reuse, then issue next load
         const auto next_iter_idx = iter_idx + kNumStages;
         if (next_iter_idx < num_iterations) {
-            ptx::tma_store_wait<kNumStages - 1>();
+            ptx::tma_store_wait();
             const auto [load_offset, num_load_bytes] = get_iter_info(next_iter_idx);
             ptx::tma_load_1d(
                 tma_buffers + stage_idx * kNumTMABytesPerStage,


polarstormx:
This has been addressed in #674 (commit 099d5f2), so closing as resolved.

PR #693, opened 2026-07-22
Pipeline direct NVLink dispatch with triple-buffered prefetch

usernamehaha2022:
## Summary

This PR pipelines direct NVLink dispatch using three shared-memory TMA buffers per dispatch warp.

Previously, a dispatch warp prepared tokens sequentially around remote S2G stores. With this change, the next token's hidden data, scale factors, metadata, and routing information are prefetched while previous remote stores remain in flight.

At steady state, the three buffers are used for:

1. A previous token whose remote S2G store may still be in flight.
2. The current token being sent to remote ranks.
3. The next token being prefetched from global memory.

Before recycling the oldest buffer, `tma_store_wait<1>()` reduces the number of outstanding store groups to one. This keeps buffer reuse safe without fully draining all remote stores between tokens.

The shared-memory calculation now accounts for all three staging buffers, and direct NVLink dispatch is limited to at most 14 dispatch warps. The non-NVLink path remains single-buffered.

## Validation

Tested on 8 x NVIDIA H100 GPUs with direct NVLink dispatch.

```bash
CUDA_VISIBLE_DEVICES=0,1,2,3,4,5,6,7 \
EP_DISABLE_GIN=1 \
python -u tests/elastic/test_ep.py \
  --num-processes 8 \
  --allow-hybrid-mode 0 \
  --test-first-only \
  --ignore-local-traffic

```

Average across 8 ranks:

| Metric | `main` | This PR | Change |
|---|---:|---:|---:|
| Dispatch | 250GB/s | 280 GB/s | 30 |
| Combine | 300 GB/s | 300 GB/s |- |

 With rank unbalanced ratio set to 1.5:

| Metric | `main` | This PR | Change |
|---|---:|---:|---:|
| Dispatch | 203GB/s | 220 GB/s | 17 |
| Combine | 209 GB/s | 209 GB/s |- |

### DIFF
--- a/csrc/kernels/elastic/dispatch.hpp
+++ b/csrc/kernels/elastic/dispatch.hpp
@@ -18,6 +18,7 @@ public:
         bool is_scaleup_nvlink;
         bool do_cpu_sync;
         bool reuse_slot_indices;
+        int num_tma_buffers;
         int num_notify_warps;
         int num_dispatch_warps; // For hybrid dispatch
         int num_scaleout_warps, num_forward_warps; // For direct dispatch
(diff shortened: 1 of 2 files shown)


alpha-baby:
my test result:

on H20 * 8 + 4 * 400 Gbps/s CX7

image

image

Issue #480, opened 2025-11-05
[bug]Missing some critical error log when timeout


huakainomore:
### 1. Description
I encountered some errors just like this. Rdma_rank 1 have this error, the sender coordinator and receiver both have timeout error.

1. For NVL receicer, `head == tail`,  it seem to be related to Forwarder.
2. For Sender coordinator, It depends on Sender to update rdma_send_channel_tail , so it seem to be related to Sender.

but both the forwarder and Sender does not, we can't know the forwarder or sender has error or not ? 
I think it's related to trap(). After trap being called by on warp, The execution of the kernel is aborted, if other warp's print buffer can't be flushed to output, it may be lost.
```
DeepEP dispatch NVL receiver timeout, channel: 1, RDMA: 1, nvl: 1, src NVL: 2, head: 75, tail: 75
DeepEP RDMA sender coordinator timeout, channel: 1, IB: 1, nvl 2, dst IB: 0, tail: 128, remaining: 21
DeepEP RDMA sender coordinator timeout, channel: 1, IB: 1, nvl 2, dst IB: 1, tail: 128, remaining: 27
DeepEP RDMA sender coordinator timeout, channel: 1, IB: 1, nvl 2, dst IB: 2, tail: 128, remaining: 30
DeepEP RDMA sender coordinator timeout, channel: 1, IB: 1, nvl 2, dst IB: 3, tail: 112, remaining: 25
DeepEP dispatch NVL receiver timeout, channel: 1, RDMA: 1, nvl: 7, src NVL: 2, head: 85, tail: 85
DeepEP dispatch NVL receiver timeout, channel: 1, RDMA: 1, nvl: 6, src NVL: 2, head: 83, tail: 83
DeepEP dispatch NVL receiver timeout, channel: 1, RDMA: 1, nvl: 5, src NVL: 2, head: 88, tail: 88
DeepEP dispatch NVL receiver timeout, channel: 1, RDMA: 1, nvl: 0, src NVL: 2, head: 76, tail: 76
DeepEP dispatch NVL receiver timeout, channel: 1, RDMA: 1, nvl: 4, src NVL: 2, head: 86, tail: 86
DeepEP dispatch NVL receiver timeout, channel: 1, RDMA: 1, nvl: 3, src NVL: 2, head: 84, tail: 84
```
### 2. Verification
[...]

PR #704, opened 2026-07-27
Make the low-latency timeout trap self-attributing in the log

KeitaW:
### Problem

CUDA 719 (`CUDA_ERROR_LAUNCH_FAILED`) is sticky: once raised on a context, every later launch on that context reports the same code. The three low-latency timeout sites in `csrc/kernels/legacy/internode_ll.cu` call `trap()` when `mask_buffer_ptr == nullptr` — that `trap()` is the **origin** of the 719 — but they print no line number and no reason, so the failure is attributed to whichever library happens to make the next launch-API call.

On a 32-rank EP32 cell the same-second 719 split **25 / 7** between DeepGEMM's `runtime_utils.hpp:143` and this file's own launch site. Both are victims; neither is the cause. Recovering the real origin took a separate `CUDA_LAUNCH_BLOCKING=1` run — the log alone could not have told anyone.

The fatal branch is not hypothetical. On a vLLM build that does not pass `enable_shrink` when constructing the low-latency `Buffer`, `buffer.py:42` defaults it to `False`, so `deep_ep.cpp:386` never allocates the mask buffer and `mask_buffer_ptr` is `nullptr` — every one of these three sites then traps rather than masks. That is the configuration this was measured on. Current vLLM main does wire it up (`all2all.py:282-284` reads `parallel_config.enable_fault_tolerance` into `enable_shrink` at `all2all.py:326`), so on main the branch is reachable whenever fault tolerance is left off, which is the default. Either way the trap is live and says nothing about where it came from.

### Change

Diagnostics only — no behaviour change. Each of the three sites (barrier, dispatch-receive, combine-receive) additionally prints:

- the resolved source line via `__LINE__`
- waited vs limit cycles
- the `mask_buffer` pointer value
- an explicit `FATAL-WILL-TRAP(origin of CUDA 719)` vs `masked-survivable` marker, selected on **the same condition that decides whether `trap()` fires**

The `if (mask_buffer_ptr == nullptr) trap();` lines themselves are untouched.

The existing `Warning: DeepEP timeout for ...` prefix is preserved verbatim so downstream log scanners keep matching. A silently blinded scanner would turn a real timeout into a false PASS, so that property is verified rather than assumed (below).

### Verification

- **`nvcc -arch=sm_90`, real H100.** The vararg lists type-check on the device path, below-threshold iterations print nothing, and `trap()` still yields exactly 719 when armed — fatal semantics unchanged.
- **Log-scanner regression.** Replayed the real device-`printf` bytes against a scanning regex: 7/7 patched lines match, 3/3 pre-patch lines match, no regressions.
- **Reachability.** Confirmed against the extracted build under test that `mask_buffer_ptr` and `enable_shrink` are both reachable, i.e. the patched branch is not dead code behind a hardcoded-false flag.

### Why `legacy/` and not the new backend

`csrc/kernels/legacy/` is currently the only DeepEP path vLLM builds against: `all2all.py:241,317` construct `deep_ep.Buffer` (`deep_ep/buffers/legacy.py`), and `ElasticBuffer` is never referenced. So this patch targets the path in production use today; it is not an argument against the direction of travel.

To be clear about what this does *not* claim: the newer trees reach `trap()` too, via `deep_ep/include/deep_ep/common/comm.cuh:46` → `ptx.cuh:21-23`, so this is not a problem that only exists in `legacy/`. The newer path is in better shape for it, though — `comm::timeout_while` (`comm.cuh:30-49`) already waits ~1 s before trapping so other threads can flush, and its timeout `printf`s already carry rank/thread/status detail (e.g. `dispatch.cuh:140`, `comm.cuh:122`). What `legacy/` is missing is exactly that: the source line and the reason, which is what this patch adds.

### Relationship to #480 and #539

#480 reports that critical timeout logs go missing. #539 adds a delay before `trap()` so other warps can flush their `printf` buffers.

This change is **complementary and non-overlapping**: #539 keeps messages from being *lost*; this one makes the message that does survive say *where it came from and why it is fatal*. Different files (`utils.cuh`/`configs.cuh` vs `internode_ll.cu`), and #539 changes `trap()`'s behaviour while this is diagnostics-only. Both are needed — with #539 alone a flushed message still carries no line number, and with this alone the message can still be lost before it reaches the host.

### DIFF
--- a/csrc/kernels/legacy/internode_ll.cu
+++ b/csrc/kernels/legacy/internode_ll.cu
@@ -58,7 +58,16 @@ __forceinline__ __device__ void barrier(int thread_id, int rank, int num_ranks, 
                 ;
             // Mask rank if timeout
             if (wait_recv_cost > LEGACY_NUM_TIMEOUT_CYCLES) {
-                printf("Warning: DeepEP timeout for barrier, rank %d, dst_rank %d\n", rank, dst_rank);
+                // SELF-ATTRIBUTION -- see the dispatch-side twin below.
+                printf("Warning: DeepEP timeout for barrier, rank %d, dst_rank %d"
+                       " [waited=%llu cycles limit=%llu mask_buffer=%p]%s internode_ll.cu:%d\n",
+                       rank,
+                       dst_rank,
+                       (unsigned long long)wait_recv_cost,
+                       (unsigned long long)LEGACY_NUM_TIMEOUT_CYCLES,
+                       (void*)mask_buffer_ptr,
+                       mask_buffer_ptr == nullptr ? " FATAL-WILL-TRAP(origin of CUDA 719)" : " masked-survivable",
+                       __LINE__);
                 if (mask_buffer_ptr == nullptr)
                     trap();
                 atomicExch(mask_buffer_ptr + dst_rank, 1);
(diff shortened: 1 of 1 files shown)


dmvevents:
Strong +1 on the framing — "both are victims; neither is the cause" matches what we hit on AWS EFA, and the sticky-719 misattribution is exactly the failure mode that cost us time.

One data point that may be useful, and a scope note: we hit a batch of `timeout (dispatch CPU)` failures across 384-expert (Kimi-K2 shape, hidden 7168, top-k 8) runs on 2× p5en over EFA. A word-boundary search for `719` / `CUDA_ERROR_LAUNCH_FAILED` / `cudaErrorLaunchFailure` across all 17 run logs returned **zero** — so those particular failures never reached the device `trap()` at all. They stalled earlier, in the **host-side** CPU busy-wait in `csrc/legacy/buffer.hpp`, where the only surviving signal is `num_recv_tokens: -1`.

So the two sites appear to be independent blind spots on the same investigation path rather than one problem:

- `csrc/kernels/legacy/internode_ll.cu` — device `trap()` → sticky 719, misattributed to the next launch (this PR)
- `csrc/legacy/buffer.hpp:596` and `:1105` — host CPU-wait throws, no source line, no elapsed-vs-limit, and no indication of *which* counter never became ready

`:596` is the starker of the two: it throws `DeepEP error: CPU recv timeout` with no diagnostics whatsoever. `:1105` prints counters via `printf`, but the exception text carries none of it, so under multi-rank log interleaving — or when only the Python traceback survives — the attribution is lost.

I've opened #708 for the host-side half, deliberately mirroring this PR's approach (diagnostics only, existing `printf` lines and error-string prefixes preserved verbatim so log scanners keep matching, plus an explicit log-scanner regression check). It touches different files, so there should be no conflict with this one.

For what it's worth, our own root cause turned out to be a build-configuration regression on our side, not a DeepEP defect — stock aws-ofi's `forced_pcie_copy()` probe requires kernel `gdrdrv >= 2.5` while our nodes carry 2.4, so the GIN gate failed deterministically and the GPU never published the counts. But the log could not have told us that: recovering it needed source archaeology plus a bisect against a known-good image. Which is precisely your argument, from the other end of the same stack.

PR #713, opened 2026-08-03
Handle missing bandwidth in SM estimation

Mr-Neutr0n:
## Problem

`ElasticBuffer.get_theoretical_num_sms` divides modeled traffic by detected RDMA and NVLink bandwidth. Both detection helpers use `0` to report that bandwidth detection failed, so a hybrid topology can raise `ZeroDivisionError` before returning an SM estimate.

## Fix

- treat traffic over an unknown-bandwidth link as unbounded transfer time when selecting the bottleneck
- retain the conservative full-device SM count when the selected bottleneck bandwidth is unavailable
- leave known-bandwidth modeling unchanged

## Regression coverage

The host-only test models a 16-rank hybrid topology with both bandwidth detectors returning `0`. Current `main` raises; this branch returns all 132 device SMs. Its fake DeepEP modules are restored after loading so the test does not pollute the wider process.

## Validation

- `python -m unittest tests.utils.test_elastic_host`
- `ruff check tests/utils/test_elastic_host.py`
- Python compilation and `git diff --check`

Prepared with OpenAI Codex assistance; I reproduced the zero-bandwidth failure and reviewed the fallback behavior.


Mr-Neutr0n:
Addressed the review: the helper now returns `0.0`, the unknown-bandwidth tie behavior is documented, and the test topology setup is centralized in one factory. https://github.com/Mr-Neutr0n/DeepEP/commit/dc9a24c

Mr-Neutr0n:
Hi @sphish - a gentle nudge on this one; it still applies and is mergeable. Happy to adjust anything, and fine to close if not wanted. Thank you.

PR #714, opened 2026-08-03
Escape generated build environment metadata

Mr-Neutr0n:
## Problem

`CustomBuildPy` interpolates persistent environment values directly into single-quoted Python literals when generating `deep_ep/envs.py`. A valid cache path containing an apostrophe produces invalid Python, while backslashes can silently change the stored value.

## Fix

Extract the source renderer and use Python `repr` for both environment names and values. Only the existing persistent-environment allowlist is emitted.

## Regression coverage

The host-only tests compile and execute the generated source, then verify exact round trips for apostrophes, backslashes, and spaces. They also verify that unrelated environment values are ignored.

## Validation

- `python -m unittest tests.utils.test_setup_helpers` (2 passed)
- `ruff check tests/utils/test_setup_helpers.py`
- Python compilation and `git diff --check`

Prepared with OpenAI Codex assistance; I reproduced the invalid generated source and reviewed the serialized output.


Mr-Neutr0n:
Addressed the review: `generate_default_envs` now reuses `build_include_dir`. I kept the renderer in `setup.py` because it is setup-only and splitting out a module solely to avoid the existing Torch build dependency would widen this fix. https://github.com/Mr-Neutr0n/DeepEP/commit/ecbce8f

Mr-Neutr0n:
Hi @sphish - a gentle nudge on this build-metadata escaping fix; still mergeable. Happy to adjust anything, and fine to close if not wanted. Thank you.

PR #715, opened 2026-08-03
fix: add system-scope release before the GIN barrier when scaleup spans NVLink and RDMA

EvadLee:
## Problem

While integrating ElasticBuffer into SGLang, multi-node **direct** mode hangs non-deterministically at the GIN barrier:

  ```
  DeepEP Gin barrier timeout, tag: 4, scaleout: 1, scaleup: 0/8, signal: 14, target: 15
  ```

Envrioment: 2 nodes x 8 H200, EP16

## Root cause

When the scaleup team spans nodes, `kIsScaleupNVLink == false` and the data plane is **mixed**: same-node peers are written by raw `ptx::tma_store_1d`, cross-node peers by `gin.put`. The GIN barrier never releases the NVLink-written half to system scope — `flush()` does not publish (`gin.h:310-311`) and `signal()` defaults to `cuda::thread_scope_device` (`gin.h:304-308`). Device scope covers RDMA, since the NIC reads *local* memory, but not stores landing in another GPU.

## Fix

`comm::gpu_barrier` already drains TMA stores before the barrier. Add a system-scope release there, on the path that publishes the raw NVLink writes:
  
  ```cpp
  if constexpr (kFlushStores) {
      ptx::tma_store_commit();
      ptx::tma_store_wait();
      __syncwarp();
      if constexpr (not kIsScaleupNVLink)
          ptx::fence_acq_rel_sys();
  }
  ```

## Validation

We could not reproduce this in a standalone unit test — it only shows up under a real multi-node workload, where the timing window is wide enough. Under SGLang on 2 nodes x 8 H200 (EP16, direct),  the barrier timeout happens every time without these two lines, and never happens with them.


sphish:
Thanks!

PR #729, opened 2026-08-16
docs: fix cached dispatch outputs

bvolpato:
## Summary

Cached decoding example discarded receiver-side routing outputs from `ElasticBuffer.dispatch`. It returned `cached_handle.topk_idx`, which stores sender-side global expert selections, and always returned `None` for weights.

## Changes

- Forward current `topk_weights` during cached dispatch.
- Return actual `recv_topk_idx` and `recv_topk_weights` outputs.

## Repro

Executed exact README code block against a contract-faithful fake buffer that returns distinct sender and receiver metadata.

Before this change:

```text
AssertionError: returned source top-k indices instead of received top-k indices; did not forward and return top-k weights
```

After this change:

```text
README cached decode example preserves dispatch outputs
```

This distinction matters in multi-rank use because sender token count and global expert IDs can differ from receiver token count and receiver-local expert IDs.

## Testing

- `uv run --isolated --with yapf==0.40.2 --with ruff==0.6.5 --with clang-format==15.0.7 -- bash ./format.sh`
- Executed exact README cached path with distinct source and received metadata before and after fix.
- Verified existing cached-dispatch test contract compares `cached_recv_topk_idx` with receiver output and exercises cached weight forwarding.

GPU suite not run. Change only corrects README example; no runtime code changed.

## Did this cause any problems?

Docs-only change. Revert this commit to restore previous example.


bvolpato:
Thanks for catching this. Agreed, the `topk_weights` docstring is contradictory. I will keep this PR focused on the README fix and address the docstring in a follow-up.

bvolpato:
Following up here. This docs-only fix is ready for review. `Code Format Check` is still awaiting maintainer approval before jobs can start.

bvolpato:
@sphish, would you mind taking a look at this? It fixes the cached `ElasticBuffer.dispatch` README example to match the receiver-side outputs and top-k weight handling in the elastic path you worked on. If someone else owns this path now, happy to redirect.

PR #744, opened 2026-09-01
[HybridEP] Support unequal per-rank token counts (ragged dispatch)

abcdabcd987:
## Summary

HybridEP required every EP rank to dispatch the same number of tokens, forcing frameworks with variable-length (no-pad) batches to pad tokens up to the group max — pad rows that ride the dispatch payload and the expert GEMMs. A line-level audit showed the requirement is not fundamental: every payload path is already per-token routing-map-gated; only the routing-map allgather size, the metadata map strides, and the chunk-grid/flag arithmetic need a *shared layout constant*, not equal real counts.

This PR lifts the restriction:

- New optional `num_of_tokens_per_rank` kwarg on `dispatch` / `dispatch_with_permute`: a group-uniform token-slot count `S` (identical on every rank, e.g. the group max of the local counts; rounded up to a multiple of 16 internally, which satisfies the TMA remainder-chunk constraint for every `ranks_per_node`). Each rank's `hidden` / `probs` / routing inputs keep their real `N_i <= S` rows. Omitted ⇒ defaults to `hidden.size(0)`, i.e. the previous uniform behavior.
- Only the **routing tensor** is padded internally to `S` rows, with routed-nowhere rows (all-`False` bool rows in sparse mode, all-`-1` sentinel rows in dense int16 top-k mode). Pad slots produce no NVLink/RDMA traffic, no permuted rows, and no expert GEMM rows — the only cost is ~`(S-N_i) × K × 2` bytes of extra routing-map allgather payload (≈0.03% of dispatch traffic at typical shapes) and zero-length chunk flags.
- The handle carries the real local count in a new trailing `num_of_valid_tokens` field (appended, so existing indices — including `handle[10]` overflow_flag — are unchanged; unpackers accept the old arity). Combine outputs are allocated at `S` (the combine kernel unconditionally writes every slot; un-routed rows come out as exact zeros) and returned narrowed to `N_i` rows.
- A zero-token rank and a group-wide-empty step both work (slot count clamped to ≥16).

## Bug fixes included

- **NIXL dispatch flag stride** used the runtime chunk count as the per-sender stride while receivers poll at the template-MAX stride — wrong slot whenever `ceil(N/chunk) != ceil(max/chunk)` on 3+ nodes (pre-existing; now MAX-based, matching DOCA).
- **NIXL signal skipped for zero-token chunks** while the dispatch receiver waits unconditionally per chunk — deadlock (pre-existing; now signals unconditionally, matching DOCA), plus a receiver-side end-of-round **drain loop** in combine so unobserved in-flight atomics can't race the residue fill when the chunk grid shrinks across calls (NIXL has no analog of DOCA's `rdma_sync` same-QP ordering).
- **NIXL prob restripe** read `num_of_tokens_per_rank` rows of the user probs tensor — OOB when slots exceed real rows; now bounded by the tensor's real row count.
- **Replaying a cached handle after buffer growth** (which reallocates buffers, resets flags, and re-registers RDMA strides) previously corrupted silently or deadlocked — now raises with a clear message in all four cached-handle paths.
- `dispatch()` now asserts routing rows == hidden rows on the fresh path (mirroring `dispatch_with_permute`), so a mismatch can't be masked by the internal padding as silent token drops.

Note: the NIXL changes are code-reviewed but compile-untested here (no NIXL build environment); DOCA and single-domain MNNVL paths are unaffected by them.

## API compatibility

- Uniform-count callers with `N % 16 == 0` (all existing tests / typical production configs) take a bit-identical path.
- Uniform-count callers with `N % 16 != 0` now get internal slot alignment + combine narrowing instead of a potential TMA failure; output shapes are unchanged (`N` rows).
- Handle tuples grew `7→8` and `11→12` by appending `num_of_valid_tokens`. Positional indices of existing fields are unchanged; fixed-arity structural unpacks of the *returned* handle need updating (one site in `tests/test_hybrid_ep.py` was).

## Validation

- **4×GB200 single node (MNNVL/IPC)**: existing `tests/test_hybrid_ep.py` (BF16+FP8, all six correctness modes) and `tests/test_graphed_hybrid_ep.py` (non-blocking + CUDA graphs) pass unchanged; new `tests/test_ragged_hybrid_ep.py` passes — bitwise vs a padded-reference `TorchRef`: sparse/dense routing × fused/non-fused permute × with/without probs × BF16/FP8, cached-handle replay bitwise (tokens, probs, scaling factors), dropped (zero-expert) tokens, zero-token ranks (incl. dense mode and FP8), and the `N % 16 != 0` default path.
- **GB200 NVL72 cluster**: same suites pass on 1 node × 4 GPUs and on 4 nodes × 16 GPUs inside a single rack segment (cross-OS-node MNNVL fabric, `USE_MNNVL=1`).
- Adversarially reviewed (multi-agent, per-finding refutation): 11 confirmed findings, all addressed in this PR.

### DIFF
--- a/csrc/hybrid_ep/hybrid_ep.cu
+++ b/csrc/hybrid_ep/hybrid_ep.cu
@@ -128,6 +128,7 @@ bool HybridEPBuffer::update_buffer(HybridEpConfigInstance config) {
     allgather_obj.update_config(buffer_config);
     release_buffer();
     allocate_buffer();
+    buffer_generation++;
   }
   return need_reallocate;
 }
@@ -136,6 +137,10 @@ HandleImpl HybridEPBuffer::metadata_preprocessing(HybridEpConfigInstance config,
   // Basic checks
   assert(local_routing_map.device().is_cuda());
   assert(local_routing_map.is_contiguous());
+  // Unreachable through the Python wrapper (which pads the routing data first),
+  // but metadata_preprocessing is a public pybind entry point: a direct caller
+  // passing mismatched rows would otherwise get silently corrupt metadata (the
+  // allgather sizes from the tensor, the scan from the scalar).
   TORCH_CHECK(local_routing_map.size(0) == num_of_tokens_per_rank,
               "local_routing_map has ", local_routing_map.size(0),
               " rows but num_of_tokens_per_rank is ", num_of_tokens_per_rank,
--- a/csrc/hybrid_ep/hybrid_ep.cu
+++ b/csrc/hybrid_ep/hybrid_ep.cu
@@ -132,10 +132,14 @@ bool HybridEPBuffer::update_buffer(HybridEpConfigInstance config) {
   return need_reallocate;
 }
 
-HandleImpl HybridEPBuffer::metadata_preprocessing(HybridEpConfigInstance config, torch::Tensor local_routing_map, int64_t num_of_tokens_per_rank, c10::optional<int64_t> num_permuted_tokens, c10::optional<int64_t> pad_multiple, bool enable_permute, bool fuse_permute_dispatch, bool non_blocking) {
+HandleImpl HybridEPBuffer::metadata_preprocessing(HybridEpConfigInstance config, torch::Tensor local_routing_map, int64_t num_of_tokens_per_rank, c10::optional<int64_t> num_of_valid_tokens, c10::optional<int64_t> num_permuted_tokens, c10::optional<int64_t> pad_multiple, bool enable_permute, bool fuse_permute_dispatch, bool non_blocking) {
   // Basic checks
   assert(local_routing_map.device().is_cuda());
   assert(local_routing_map.is_contiguous());
+  TORCH_CHECK(local_routing_map.size(0) == num_of_tokens_per_rank,
+              "local_routing_map has ", local_routing_map.size(0),
+              " rows but num_of_tokens_per_rank is ", num_of_tokens_per_rank,
+              "; pad the routing data to the group-uniform token-slot count");
   if(fuse_permute_dispatch) {
     assert(enable_permute);
   }
@@ -147,11 +151,12 @@ HandleImpl HybridEPBuffer::metadata_preprocessing(HybridEpConfigInstance config,
 
   // Run the hybrid-ep metadata preprocessing kernel
   auto handle = executor.metadata_preprocess_core(
-    config, 
-    nvl_coordinator.preprocessing_tmp, 
+    config,
+    nvl_coordinator.preprocessing_tmp,
     nvl_coordinator.preprocessing_local_experts_tmp,
-    global_routing_map, 
-    num_of_tokens_per_rank, 
+    global_routing_map,
+    num_of_tokens_per_rank,
+    num_of_valid_tokens.has_value() ? num_of_valid_tokens.value() : -1,
     num_permuted_tokens.has_value() ? num_permuted_tokens.value() : -1,
     pad_multiple.has_value() ? pad_multiple.value() : 0,
     enable_permute,
@@ -233,7 +238,11 @@ HybridEPBuffer::combine(
            probs.value().size(1) == config.num_of_experts_per_rank * config.num_of_ranks_per_node);
   }
 
-  // Construct the output tensors
+  // Construct the output tensors. The combine kernel writes every token slot
+  // in [0, num_of_tokens_per_rank), so the buffers are allocated at the
+  // group-uniform slot count; the returned views are narrowed to the real
+  // local token count.
+  const int64_t num_of_valid_tokens = handle.valid_token_count();
   torch::Tensor combined_tokens, combined_probs;
   combined_tokens =torch::empty({handle.num_of_tokens_per_rank, config.hidden_dim},
                    torch::dtype(hidden.dtype()).device(torch::kCUDA));
@@ -260,7 +269,11 @@ HybridEPBuffer::combine(
   executor.combine_preprocess(config, args);
   executor.combine_core(config, args);
   executor.combine_postprocess(config, args);
-  
+
+  if (num_of_valid_tokens != handle.num_of_tokens_per_rank) {
+    combined_tokens = combined_tokens.narrow(0, 0, num_of_valid_tokens);
+    if (with_probs) combined_probs = combined_probs.narrow(0, 0, num_of_valid_tokens);
+  }
   return std::make_tuple(combined_tokens, combined_probs);
 }
 
@@ -356,7 +369,10 @@ HybridEPBuffer::combine_with_unpermute(
     assert(probs.value().dtype() == torch::kFloat32);
   }
 
-  // Construct the output tensors
+  // Construct the output tensors. See combine(): allocate at the group-uniform
+  // slot count (the kernel writes every slot), return views narrowed to the
+  // real local token count.
+  const int64_t num_of_valid_tokens = handle.valid_token_count();
   torch::Tensor combined_tokens, combined_probs;
   combined_tokens =torch::empty({handle.num_of_tokens_per_rank, config.hidden_dim},
                    torch::dtype(hidden.dtype()).device(torch::kCUDA));
@@ -386,6 +402,10 @@ HybridEPBuffer::combine_with_unpermute(
   executor.combine_preprocess(config, args);
   executor.combine_core(config, args);
   executor.combine_postprocess(config, args);
-  
+
+  if (num_of_valid_tokens != handle.num_of_tokens_per_rank) {
+    combined_tokens = combined_tokens.narrow(0, 0, num_of_valid_tokens);
+    if (with_probs) combined_probs = combined_probs.narrow(0, 0, num_of_valid_tokens);
+  }
   return std::make_tuple(combined_tokens, combined_probs);
 }
--- a/csrc/hybrid_ep/hybrid_ep.cuh
+++ b/csrc/hybrid_ep/hybrid_ep.cuh
@@ -26,6 +26,11 @@ public:
   ~HybridEPBuffer();
   bool update_buffer(HybridEpConfigInstance config); // True means the buffer is reallocated.
 
+  // Incremented on every buffer reallocation. Handles record the generation
+  // they were created against (via HybridEpConfigInstance::buffer_generation)
+  // so stale-handle replay can be rejected.
+  int64_t buffer_generation = 0;
+
   HandleImpl metadata_preprocessing(HybridEpConfigInstance config,
     torch::Tensor local_routing_map,
     int64_t num_of_tokens_per_rank,
--- a/csrc/hybrid_ep/hybrid_ep.cuh
+++ b/csrc/hybrid_ep/hybrid_ep.cuh
@@ -28,8 +28,9 @@ public:
 
   // Incremented on every buffer reallocation. Handles record the generation
   // they were created against (via HybridEpConfigInstance::buffer_generation)
-  // so stale-handle replay can be rejected.
-  int64_t buffer_generation = 0;
+  // so stale-handle replay can be rejected. Starts at 1 so a config that was
+  // never stamped (generation 0) can never match a live buffer.
+  uint64_t buffer_generation = 1;
 
   HandleImpl metadata_preprocessing(HybridEpConfigInstance config,
     torch::Tensor local_routing_map,
--- a/csrc/hybrid_ep/hybrid_ep.cuh
+++ b/csrc/hybrid_ep/hybrid_ep.cuh
@@ -26,12 +26,13 @@ public:
   ~HybridEPBuffer();
   bool update_buffer(HybridEpConfigInstance config); // True means the buffer is reallocated.
 
-  HandleImpl metadata_preprocessing(HybridEpConfigInstance config, 
-    torch::Tensor local_routing_map, 
-    int64_t num_of_tokens_per_rank, 
-    c10::optional<int64_t> num_permuted_tokens, 
-    c10::optional<int64_t> pad_multiple, 
-    bool enable_permute, 
+  HandleImpl metadata_preprocessing(HybridEpConfigInstance config,
+    torch::Tensor local_routing_map,
+    int64_t num_of_tokens_per_rank,
+    c10::optional<int64_t> num_of_valid_tokens,
+    c10::optional<int64_t> num_permuted_tokens,
+    c10::optional<int64_t> pad_multiple,
+    bool enable_permute,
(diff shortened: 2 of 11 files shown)


tests/test_hybrid_ep.py
```diff
@@ -182,7 +182,7 @@ def test_hybrid_ep_correctness(buffer: deep_ep.HybridEPBuffer, ref: TorchRef, us
                 masked_probs[:, start:end] = dispatched_probs_dense[:, start:end]
                 dispatched_probs_dense = masked_probs
 
-            _, _, _, num_dispatched_tokens, local_expert_routing_map, _, _ = handle_dense
```
Autumn1998, line 185: why we need this change🤔  I’d prefer to keep the changes to a minimum.

deep_ep/hybrid_ep_buffer.py
```diff
@@ -263,6 +332,36 @@ def update_template_config(
         self.runtime.update_buffer(config)
         return config
 
+    @staticmethod
+    def _unpack_dispatch_handle(handle_impl, handle: tuple):
```
Autumn1998: I’d probably not using this kind of pack/unpack helper, since the `handle` for `dispatch`/`combine` and `unpermute`/`permute` are all different. The helper would have limited reuse and would add mental overhead when reading the code. Additionally, I’d like to keep the PR changes as minimal as possible.

deep_ep/hybrid_ep_buffer.py
```diff
@@ -263,6 +332,36 @@ def update_template_config(
...
+            handle_impl.local_expert_routing_map,
+            handle_impl.num_of_tokens_per_rank,
+            handle_impl.config,
+        ) = handle[:7]
+        handle_impl.num_of_valid_tokens = (
+            handle[7] if len(handle) > 7 else handle_impl.num_of_tokens_per_rank
+        )
+
+    @staticmethod
+    def _pack_dispatch_handle(handle_impl):
```
Autumn1998: ditto

deep_ep/hybrid_ep_buffer.py
```diff
@@ -366,26 +469,60 @@ def combine(
...
+
+        combined_token, combined_probs = self.runtime.combine(
+            hidden=hidden,
+            probs=probs,
+            handle=handle_impl,
+            with_probs=probs is not None,
+        )
+        return combined_token, combined_probs
+
+    @staticmethod
```
Autumn1998: ditto

deep_ep/hybrid_ep_buffer.py
```diff
@@ -366,26 +469,60 @@ def combine(
...
+            handle[11] if len(handle) > 11 else handle_impl.num_of_tokens_per_rank
+        )
 
-        combined_token, combined_probs = self.runtime.combine(
-            hidden=hidden,
-            probs=probs,
-            handle=handle_impl,
-            with_probs=probs is not None,
+    @staticmethod
+    def _pack_permute_handle(handle_impl):
```
Autumn1998: ditto

deep_ep/hybrid_ep_buffer.py
```diff
@@ -161,6 +161,75 @@ def _use_dense_topk_routing(self, num_of_experts: int, num_of_experts_per_rank:
...
+            num_of_tokens_per_rank = num_of_valid_tokens
+        num_of_tokens_per_rank = int(num_of_tokens_per_rank)
+        assert num_of_tokens_per_rank >= num_of_valid_tokens, (
+            f"num_of_tokens_per_rank ({num_of_tokens_per_rank}) must be >= the local "
+            f"token count ({num_of_valid_tokens})"
+        )
+        num_of_tokens_per_rank = max(num_of_tokens_per_rank, 16)
+        return (num_of_tokens_per_rank + 15) // 16 * 16
+
+    def _check_handle_buffer_compat(self, handle_impl):
```
Autumn1998, line 191: I suggest removing this check, because an unchanged `max_num_of_tokens_per_rank` does not necessarily mean that the buffer has not been reused.

csrc/hybrid_ep/hybrid_ep.cu
```diff
@@ -132,10 +132,14 @@ bool HybridEPBuffer::update_buffer(HybridEpConfigInstance config) {
   return need_reallocate;
 }
 
-HandleImpl HybridEPBuffer::metadata_preprocessing(HybridEpConfigInstance config, torch::Tensor local_routing_map, int64_t num_of_tokens_per_rank, c10::optional<int64_t> num_permuted_tokens, c10::optional<int64_t> pad_multiple, bool enable_permute, bool fuse_permute_dispatch, bool non_blocking) {
+HandleImpl HybridEPBuffer::metadata_preprocessing(HybridEpConfigInstance config, torch::Tensor local_routing_map, int64_t num_of_tokens_per_rank, c10::optional<int64_t> num_of_valid_tokens, c10::optional<int64_t> num_permuted_tokens, c10::optional<int64_t> pad_multiple, bool enable_permute, bool fuse_permute_dispatch, bool non_blocking) {
   // Basic checks
   assert(local_routing_map.device().is_cuda());
   assert(local_routing_map.is_contiguous());
+  TORCH_CHECK(local_routing_map.size(0) == num_of_tokens_per_rank,
```
Autumn1998, line 144: This check looks like it can never fail, since the routing map has already been padded to `num_of_tokens_per_rank`.

deep_ep/hybrid_ep_buffer.py
abcdabcd987: Fair point — done in 0aa8753: the four helpers are inlined at their call sites. One property I kept from the helper version: the unpacks are strict fixed-arity (8 targets for the dispatch handle, 12 for the permute handle), so passing a wrong-kind or truncated handle fails immediately with a ValueError instead of silently misassigning fields.

deep_ep/hybrid_ep_buffer.py
abcdabcd987: Done in 0aa8753 (inlined).

deep_ep/hybrid_ep_buffer.py
abcdabcd987: Done in 0aa8753 (inlined).

deep_ep/hybrid_ep_buffer.py
abcdabcd987: Done in 0aa8753 (inlined).

tests/test_hybrid_ep.py
abcdabcd987: This change is forced by the handle layout: the dispatch handle gained a trailing num_of_valid_tokens element (7 → 8), so the old fixed 7-target unpack raises `ValueError: too many values to unpack`. Switched to indexed access so the line is also robust to any future appended fields. (The field is what lets combine return exactly hidden.size(0) rows when ranks dispatch unequal counts; it is appended at the end so all existing tuple indices are unchanged.)

deep_ep/hybrid_ep_buffer.py
abcdabcd987: You're right that the max_num_of_tokens_per_rank comparison was incomplete — reallocation can also be triggered by growth in hidden_dim / experts-per-rank / chunk-size changes etc., and the check would pass in those cases. Rather than dropping the protection (replaying a handle across any reallocation silently corrupts the flag protocol or deadlocks, since flags are reset and buffers re-registered), 0aa8753 replaces it with a buffer generation counter: HybridEPBuffer::buffer_generation increments on every release/reallocate cycle and is stamped into the config each handle carries, so replay after any reallocation — whatever the trigger — is rejected with a clear error. Added a test that grows the buffer mid-run and verifies both the rejection and post-growth correctness. Happy to split this guard into a separate PR if you'd prefer this one stay minimal.

csrc/hybrid_ep/hybrid_ep.cu
abcdabcd987: Correct that it's unreachable through the Python wrapper (which pads the routing data first). I kept it because metadata_preprocessing is a public pybind entry point: a direct caller passing routing rows != num_of_tokens_per_rank previously got silently corrupt metadata (the allgather sizes from the tensor while the scan kernel runs with the scalar). 0aa8753 adds a comment stating exactly that rationale. If you'd still prefer it dropped, no objection — it's purely boundary defense.

tests/test_ragged_hybrid_ep.py
```diff
@@ -0,0 +1,513 @@
...
+    else:
+        hidden = torch.randn(num_valid, hidden_dim, device="cuda", dtype=torch.bfloat16, generator=gen)
+    scaling_factor = torch.randn(num_valid, hidden_dim // 128, device="cuda", dtype=torch.float32, generator=gen)
+
+    probs = torch.zeros(num_valid, num_of_experts, device="cuda", dtype=torch.float32)
+    routing_map = torch.zeros(num_valid, num_of_experts, device="cuda", dtype=torch.bool)
+    topk_idx = torch.full((num_valid, topk), -1, device="cuda", dtype=torch.int64)
+    topk_weights = torch.zeros(num_valid, topk, device="cuda", dtype=torch.float32)
+
+    dropped = torch.rand(num_valid, device="cuda", generator=gen) < drop_fraction
```
Autumn1998: It seems that `num_valid` already simulates the case where each rank has a different number of tokens. Would adding this drop behavior be somewhat redundant? Do we need it?

csrc/hybrid_ep/hybrid_ep.cuh
```diff
@@ -26,12 +26,18 @@ public:
...
-  HandleImpl metadata_preprocessing(HybridEpConfigInstance config, 
-    torch::Tensor local_routing_map, 
-    int64_t num_of_tokens_per_rank, 
-    c10::optional<int64_t> num_permuted_tokens, 
-    c10::optional<int64_t> pad_multiple, 
-    bool enable_permute, 
+  // Incremented on every buffer reallocation. Handles record the generation
+  // they were created against (via HybridEpConfigInstance::buffer_generation)
+  // so stale-handle replay can be rejected.
+  int64_t buffer_generation = 0;
```
Autumn1998: Maybe uint64_t will be better?

abcdabcd987: Done in 2364852 — both counters are uint64_t now. While changing the type I also tightened the sentinel semantics: the buffer's generation starts at 1 and the config field defaults to 0 ("never stamped"), so a handle whose config never went through the wrapper can't accidentally match a live buffer (previously that role was played by the -1 default, which is what motivated the signed type).

tests/test_ragged_hybrid_ep.py
abcdabcd987: They cover different behavior, so I'd keep both:

1. **Different scan pattern.** The pad rows the library appends are a contiguous all-false *suffix* at the end of the slot range. The dropped rows here are *interleaved holes inside the valid region*, so the metadata scan's dense-slot compaction has to skip gaps between real tokens rather than just stop early — a different access pattern through the prefix-sum/compaction logic.

2. **Different observability.** Pad rows are trimmed from the combine output, so their zero-fill behavior is never user-visible. Dropped rows sit inside the returned region, and the test asserts they combine to *exact zeros* — that's the documented "-1 sentinel / all-false row" user contract from the dispatch docstring, which no other test covers (the pre-existing tests route every token to exactly topk experts).

3. The internal routing-pad mechanism is built on exactly these dropped-row semantics, so pinning the contract with user-visible assertions guards the foundation the ragged feature rests on.

That said, if you'd prefer the ragged cases stay pure, I'm happy to set DROP_FRACTION to 0 in the main matrix and move the dropped-token coverage into its own dedicated case.

QiaoK:
Looks good to me.

Verified on 2× GB200 nodes (4 GPUs each, EP=8) in `nvcr.io/nvidia/pytorch:26.07-py3`,
with hidden=7168, 8 local experts, top-k 8, 4096 tokens/rank, bf16 and FP8:

| test | 1 node | 2 nodes |
|---|---|---|
| `test_hybrid_ep.py` | pass | pass |
| `test_ragged_hybrid_ep.py` | pass | pass |
| `test_graphed_hybrid_ep.py` | pass | pass |

QiaoK (approved):

tests/test_ragged_hybrid_ep.py
Autumn1998: Yes, but in the training scenario, we should only be padding at the end. I don’t think we need to over-simulate this case, and keeping it simpler would make the code cleaner as well. What do you think?

abcdabcd987: Agreed — done in 268ca93: the ragged matrix is now drop-free (padding only at the end, matching the training scenario). The documented -1 topk sentinel / all-false-row contract keeps one compact dedicated case (test_dropped_tokens: sparse + dense, dispatch+combine, asserting zero-expert rows produce exact-zero output), since no other test exercises it. Main-matrix code is simpler as you suggested.

abcdabcd987:
@QiaoK thanks for the independent verification — the 2-node runs also close the multi-node coverage gap flagged in the PR description for the MNNVL path.

Latest state: all of @Autumn1998's comments are addressed as of 268ca93 — helpers inlined, buffer-generation guard (uint64_t), and the ragged test matrix is now end-padded only, with the -1 sentinel contract kept in one dedicated dropped-token case. Full suite re-verified locally on 4×GB200 (uniform + ragged + graphed, BF16+FP8).

tests/test_ragged_hybrid_ep.py
Autumn1998: What I meant was to remove the dropped-token-related logic entirely, including dropped, drop_fraction, and tests such as test_dropped_tokens. https://github.com/deepseek-ai/DeepEP/commit/268ca934a575b9b967af944b81feaf148489f343 make the pr more complex.

abcdabcd987: Done in 7c646d6 — removed entirely: `dropped`, `drop_fraction`/`DROP_FRACTION`, `test_dropped_tokens`, and the zero-row assertion in `check_combined`. Every valid row in the ragged tests now routes to exactly topk experts, so the only all-false rows are the end-padding the library adds, matching the training scenario. Suite re-verified locally (4×GB200, BF16+FP8, all pass).

Autumn1998 (approved):
LGTM overall. cc @jershi425  for final review

abcdabcd987:
Thanks for the review! @Autumn1998 

@jershi425 Would you mind taking a look once you got time? Thanks!

jershi425:
> Thanks for the review! @Autumn1998
> 
> @jershi425 Would you mind taking a look once you got time? Thanks!

Merged, thanks for the contribution.

PR #751, opened 2026-09-14
Ensure that the last get rings the doorbell

Yiltan:


deep_ep/include/deep_ep/impls/engram_fetch.cuh
```diff
@@ -67,10 +83,13 @@ engram_fetch_impl(const ncclDevComm_t nccl_dev_comm, const ncclWindow_t nccl_win
 
             // Issue RDMA get
             const auto request_idx = atomicAdd_block(num_requests_per_peer + peer_idx, 1);
+            const bool is_last_get = (request_idx == total_requests_per_peer[peer_idx] - 1);
```
Yiltan, line 85: Alternative solution could be to use the loop condition to determine the last get op

polarstormx:
This has been addressed in #752 (commit a56d615), so closing as superseded.

